PR #15, opened 2020-11-12
Add docu about CAP and SDK integration

Emdee89:
This documentation illustrates how the Cloud SDK can be integrated into an existing CAP app.

### DIFF
--- a/docs/java/guides/cap-sdk-integration.mdx
+++ b/docs/java/guides/cap-sdk-integration.mdx
@@ -70,7 +70,7 @@ Note that the `version` tags can be omitted, as the Cloud SDK BOM you added befo
 That is, you only need to increase the version of the BOM and all modules get updated automatically.
 :::
 
-:::note 
+:::tip
 The two mentioned dependencies are modules which cover a broad range of possible Cloud SDK use cases.
 You may choose a more fine-grained selection of Cloud SDK dependencies according to your specific use case.
 :::
@@ -98,5 +98,4 @@ We'll annotate the Spring Boot application class which is annotated with `@Sprin
 Now the Cloud SDK is integrated into with the CAP application.
 You can go ahead using the Cloud SDK features, such as querying OData services.
 :::
-
 
--- a/docs/java/guides/cap-sdk-integration.mdx
+++ b/docs/java/guides/cap-sdk-integration.mdx
@@ -1,0 +1,85 @@
+---
+id: cap-sdk-integration
+title: Integration with Cloud Application Programming (CAP) Model
+hide_title: false
+hide_table_of_contents: false
+sidebar_label: Integration with CAP
+description: Integrate the SAP Cloud SDK into your CAP application
+keywords:
+- sap
+- cloud
+- sdk
+- cloud native
+- cloud sdk
+- sap cloud sdk
+- cap
+image:
+---
+
+This document outlines how you integrate the SAP Cloud SDK into an application based on the [SAP Cloud Application Programming Model](https://cap.cloud.sap/docs/) (referred to as *CAP* in the remainder).
+
+:::success Check out the tutorial as well
+While this document provides a general overview on this topic, we have published a practical [tutorial on SAP Developers](https://developers.sap.com/tutorials/cloudsdk-integrate-cap.html) which illustrates all steps on a comprehensive example. 
+:::
+
+## Integration Steps
+
+:::note Assumption
+The following instructions assume that you have an existing CAP project that can be built successfully.
+Furthermore, we assume that you deploy your app to the SCP Cloud Foundry environment.
+:::
+
+### Add the SAP Cloud SDK BOM
+Add the SAP Cloud SDK Bill-of-Material (BOM) into the `dependencyManagement` section of the root POM.
+Here is the XML snippet:
+```xml
+<dependencyManagement>
+    <dependencies>
+        <dependency>
+            <groupId>com.sap.cloud.sdk</groupId>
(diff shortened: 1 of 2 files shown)


docs/java/guides/cap-sdk-integration.mdx
```diff
@@ -0,0 +1,85 @@
...
+    </dependencies>        
+</dependencyManagement>
+```
+:::success Use latest SDK version
+We recommend using the **latest** SAP Cloud SDK version always. 
+Check out the [release notes](../release-notes-sap-cloud-sdk-for-java) to see the release history.
+:::
+
+### Add the SAP Cloud SDK Dependencies
+Add the respective Cloud SDK dependencies to the POM file under the `srv` directory.
```
sacnl, line 54: Add a statement that the concrete dependencies to add may be different depending on requirements. (Do we have docs that explain which module to use for what for minimal footprint?)

docs/java/guides/cap-sdk-integration.mdx
```diff
@@ -0,0 +1,85 @@
...
+    <artifactId>s4hana-all</artifactId>
+</dependency>
+```
+:::note Version information controlled through Cloud SDK BOM
+Note that the `version` tags can be omitted, as the Cloud SDK BOM you added beforehand controls the version of all SDK modules.
+That is, you only need to increase the version of the BOM and all modules get updated automatically.
+:::
+
+### Add the Integration Dependency
+In order to integrate the tenant and user concepts between CAP and the Cloud SDK, we'll add a respective integration dependency to the POM file under the `srv` directory.
```
sacnl, line 72: Do I always have to do that? If not, in which cases? Mention that this goes into the same place as before.

docs/java/guides/cap-sdk-integration.mdx
```diff
@@ -0,0 +1,85 @@
...
+- cloud
+- sdk
+- cloud native
+- cloud sdk
+- sap cloud sdk
+- cap
+image:
+---
+
+This document outlines how you integrate the SAP Cloud SDK into an application based on the [SAP Cloud Application Programming Model](https://cap.cloud.sap/docs/) (referred to as *CAP* in the remainder).
```
sacnl, line 19: Mention in how far this differs from adding the SDK to any other Java project. (Do we have documentation for that?)

sacnl (changes requested):
Component scan annotation is not required?

docs/java/guides/cap-sdk-integration.mdx
Emdee89: I find it difficult to mention here at each step "This is relevant for CAP only" or so. I would say we create a document about integration "with any other Java project" in the future and let this CAP integration docu only mention the additional steps.

Meaning, I would leave this mention out here for now.

docs/java/guides/cap-sdk-integration.mdx
Emdee89: I added a note on this topic.

docs/java/guides/cap-sdk-integration.mdx
Emdee89: Further explanation added

Emdee89:
> Component scan annotation is not required?

I added a section about this.

docs/java/guides/cap-sdk-integration.mdx
```diff
@@ -0,0 +1,102 @@
...
+    <groupId>com.sap.cloud.sdk.s4hana</groupId>
+    <artifactId>s4hana-all</artifactId>
+</dependency>
+```
+:::note Version information controlled through Cloud SDK BOM
+Note that the `version` tags can be omitted, as the Cloud SDK BOM you added beforehand controls the version of all SDK modules.
+That is, you only need to increase the version of the BOM and all modules get updated automatically.
+:::
+
+:::note
```
artemkovalyov, line 73:
```suggestion
:::tip
```

docs/java/guides/cap-sdk-integration.mdx
```diff
@@ -0,0 +1,102 @@
...
+```
+:::note Version information controlled through Cloud SDK BOM
+Note that the `version` tags can be omitted, as the Cloud SDK BOM you added beforehand controls the version of all SDK modules.
+That is, you only need to increase the version of the BOM and all modules get updated automatically.
+:::
+
+:::note 
+The two mentioned dependencies are modules which cover a broad range of possible Cloud SDK use cases.
+You may choose a more fine-grained selection of Cloud SDK dependencies according to your specific use case.
+:::
```
artemkovalyov, line 76: I think we need at least one specific example. Otherwise, this sounds very broad to me. Let's at least mention some specific use-case and give module names to use.

artemkovalyov (approved):
One obvious example of fine-grained dependencies would be nice. Give just use-case and name of module - no need to provide another extract from the pom.

artemkovalyov (changes requested):
Sorry, meant to ask for a change with the last one. Request me again when done.

docs/java/guides/cap-sdk-integration.mdx
Emdee89: Let's create a new dedicated page for that in the future.

artemkovalyov (approved):
Thanks, Marco, let's not forget a BLI about integrating SDK into an existing project.

artemkovalyov:
@sacnl, take a look if you're ok with changes/suggestions to unblock the merge?

sacnl (approved):

PR #16, opened 2020-11-16
Remove superfluous section from CAP docu

Emdee89:
Please provide a description of what your change does and why it is needed.


artemkovalyov (approved):
Funnily slipped through. Merging it.

PR #29, opened 2020-11-23
Add warning about setting the service path for geneterated OData classes

MatKuhr:
This PR adds a warning to override the service path if the generated default is not matching expectations.

based on customer feedback e.g. https://stackoverflow.com/questions/64807793/unable-to-use-vdm-generator-with-cap-for-gwsample-basic-service

### DIFF
--- a/docs/java/features/odata/generate-client.mdx
+++ b/docs/java/features/odata/generate-client.mdx
@@ -166,8 +166,8 @@ Adapt the `<inputDirectory>` to point to the location of your service definition
 Now maven will run the generator within the `process-sources` phase which is executed before compile.
 Also the `<compileScope>` option instructs maven to add the generated code as sources for the compile phase.
 
-:::note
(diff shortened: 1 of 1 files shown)


docs/java/features/odata/generate-client.mdx
```diff
@@ -166,8 +166,8 @@ Adapt the `<inputDirectory>` to point to the location of your service definition
 Now maven will run the generator within the `process-sources` phase which is executed before compile.
 Also the `<compileScope>` option instructs maven to add the generated code as sources for the compile phase.
 
-:::note
-Please note that if you use the generator for services other than SAP S/4HANA services, you need to add the parameter `defaultBasePath` to the configuration section, which should provide the base path to the exposed API (e.g _odata/v4/_).
```
MatKuhr, line 170: This note was incorrect. Also users will pretty much always use it for something other than S/4 services. Because the SDK already ships the generated code for that.

artemkovalyov (approved):
LGTM. If this note is often needed we can convert it into a component:)

PR #155, opened 2021-03-05
Add execute raw

jjtang1985:
## What Has Changed?

Add doc for `executeRaw`.

## Manual Checks?

- [ ] You have updated the [feature matrix](https://github.com/SAP/cloud-sdk/blob/main/docs/components/data/features.js) if you add documentation on a new feature or otherwise required.


docs/js/features/odata/execute-odata-request.mdx
```diff
@@ -188,4 +188,19 @@ MyEntity.requestBuilder()
 ```
 
 This will add the custom service path to your request.
-Executing the example request above against a destination with URL 'https://my.s4-system.com' would then execute the request against something like this: 'https://my.s4-system.com/my/custom/service/path/MyEntity'.
+Executing the example request above against a destination with URL `https://my.s4-system.com` would then execute the request against something like this: `https://my.s4-system.com/my/custom/service/path/MyEntity`.
+
+### Getting the Raw Response
+
+There are some valid use cases that the typed response does not meet special requirements like below:
```
artemkovalyov, line 195:
```suggestion
There are some valid use cases with requirements that a typed response doesn't meet:
```

docs/js/features/odata/execute-odata-request.mdx
```diff
@@ -188,4 +188,19 @@ MyEntity.requestBuilder()
...
+
+- the response cannot be deserialized automatically, so the raw response can work as workaround
+- instead of the response body, additional information is needed like status code or response header
+- in addition to the response, the request built by the SAP Cloud SDK needs also to be returned
```
artemkovalyov, line 199:
```suggestion
- the response cannot be deserialized automatically, so the raw response can work as a workaround
- instead of the response body, additional information is needed like status code or response header
- in addition to the response, the request built by the SAP Cloud SDK needs also to be returned
```

docs/js/features/odata/execute-odata-request.mdx
```diff
@@ -188,4 +188,19 @@ MyEntity.requestBuilder()
...
+
+To fulfill the requirement shown above, the SAP Cloud SDK provides the `executeRaw` function that returns the original request and the corresponding raw response.
+
+:::note
```
artemkovalyov, line 203:
```suggestion
:::note typed response from `execute()` is the best practice
```

docs/js/features/odata/execute-odata-request.mdx
```diff
@@ -188,4 +188,19 @@ MyEntity.requestBuilder()
...
+In general, to make a request, using `execute` function for returning a typed response is the best practise.
+The `executeRaw` should only be used for good reasons.
```
artemkovalyov, line 205:
```suggestion
Use `executeRaw` only for workarounds, debugging, logging, and other good reasons where the typed `execute` method doesn't provide enough flexibility.
```

artemkovalyov (changes requested):
I suggest fixes for better readability.

artemkovalyov (approved):
LGTM, unless you want reviews from Marika and Frank before merging.

docs/js/features/odata/execute-odata-request.mdx
```diff
@@ -188,4 +188,18 @@ MyEntity.requestBuilder()
...
 
 This will add the custom service path to your request.
-Executing the example request above against a destination with URL 'https://my.s4-system.com' would then execute the request against something like this: 'https://my.s4-system.com/my/custom/service/path/MyEntity'.
+Executing the example request above against a destination with URL `https://my.s4-system.com` would then execute the request against something like this: `https://my.s4-system.com/my/custom/service/path/MyEntity`.
+
+### Getting the Raw Response
+
+There are some valid use cases with requirements that a typed response doesn't meet:
+
+- the response cannot be deserialized automatically, so the raw response can work as a workaround
```
marikaner, line 197: I would like to add an example to this point when this might be the case .

docs/js/features/odata/execute-odata-request.mdx
```diff
@@ -188,4 +188,18 @@ MyEntity.requestBuilder()
...
+- instead of the response body, additional information is needed like status code or response header
+- in addition to the response, the request built by the SAP Cloud SDK needs also to be returned
```
marikaner, line 199: I am really sorry for not reviewing your PR on this. I wasn't aware that we also return the ODataRequest instance that we create and I am not sure that this is the best idea. Let's discuss this offline.

docs/js/features/odata/execute-odata-request.mdx
```diff
@@ -188,4 +188,18 @@ MyEntity.requestBuilder()
...
+
+To fulfill the requirement shown above, the SAP Cloud SDK provides the `executeRaw` function that returns the original request and the corresponding raw response.
```
marikaner, line 201: "the requirement(s) shown above" is quite abstract. I would try to tell from the beginning what this is about and than keep the name up. Maybe:
```suggestion
To retrieve a raw response for your request use the `executeRaw` method.
```

docs/js/features/odata/execute-odata-request.mdx
```diff
@@ -188,4 +188,18 @@ MyEntity.requestBuilder()
...
+
+:::note typed response from `execute()` is the best practice
```
marikaner, line 203: This should be a title and uppercase
```suggestion
:::note Best Practice: Retrieve typed responses using `execute()`
```

docs/js/features/odata/execute-odata-request.mdx
```diff
@@ -188,4 +188,18 @@ MyEntity.requestBuilder()
...
+Use `executeRaw` only for workarounds, debugging, logging, and other good reasons where the typed `execute` method doesn't provide enough flexibility.
```
marikaner, line 204: Do you think that this is true? Is there anything wrong with `executeRaw` so that we need to speak out this warning?

marikaner (changes requested):
Looks mostly good, but I am missing an example.

docs/js/features/odata/execute-odata-request.mdx
```diff
@@ -188,4 +188,18 @@ MyEntity.requestBuilder()
 ```
 
 This will add the custom service path to your request.
-Executing the example request above against a destination with URL 'https://my.s4-system.com' would then execute the request against something like this: 'https://my.s4-system.com/my/custom/service/path/MyEntity'.
+Executing the example request above against a destination with URL `https://my.s4-system.com` would then execute the request against something like this: `https://my.s4-system.com/my/custom/service/path/MyEntity`.
+
+### Getting the Raw Response
```
FrankEssenberger, line 193: Perhaps I would include also the Request like `Getting the Raw Request/Response`

docs/js/features/odata/execute-odata-request.mdx
jjtang1985: This is rather a guidance/recommendation instead of a warning.
Here are some reasons.
- "type-safe" is one of the keywords of the sdk. `execute` gives you a deserialised typed object instead of a raw json. Here are some pros of the typed object:
  - No need for the user to care the response structure. Consider the difference of `get all response` and `get by key response`.
  - No need for the user to deserialise a json, which might be used in e.g., an `update request builder`.
  - IDE auto-completion when you want to get a property value
- For the current main branch (might need to be discussed), the return type contains the `HttpRequest` and the `HttpResponse` (`body` + `header` + ...), which is far more than the typed object from `execute`. For some big entities and consider batch, this might lead to unnecessary performance issue.

marikaner: I see your point, that in most cases we recommend to use the deserialized types. But there are use cases where you might *need* `executeRaw` - e. g. if you want to access the status code. This is not a workaround (or debugging or logging).
This side note sounds like there might be something wrong with `executeRaw`.

Maybe to make it sound less harsh, instead of putting this in a separate note here, you could lead with that, something in that direction: "For most cases we recommend retrieving the deserialized response using `execute`. In some cases..."

docs/js/features/odata/execute-odata-request.mdx
```diff
@@ -188,4 +188,33 @@ MyEntity.requestBuilder()
...
 
 This will add the custom service path to your request.
-Executing the example request above against a destination with URL 'https://my.s4-system.com' would then execute the request against something like this: 'https://my.s4-system.com/my/custom/service/path/MyEntity'.
+Executing the example request above against a destination with URL `https://my.s4-system.com` would then execute the request against something like this: `https://my.s4-system.com/my/custom/service/path/MyEntity`.
+
+### Getting the Raw Response and the Original Request
+
+In most cases, it is recommended to use `execute` method to retrieve the deserialized typed response.
+
+In some cases, on the other hand, you might prefer to retrieve the raw response and the original request:
```
marikaner, line 197: My english teacher always said you cannot use "on the other hand" if you didn't use "on the one hand". ;)
```suggestion
In some cases, you might prefer to retrieve the raw response and the original request:
```

marikaner (approved):
One minor suggestion, otherwise LGTM

docs/js/features/odata/execute-odata-request.mdx
jjtang1985: Here is one [answer](https://english.stackexchange.com/questions/352529/can-on-the-other-hand-be-used-without-on-one-hand) with the reference from Cambridge dictionary. But I'm fine without `on the other hand`, which seems to be formal usage.

PR #183, opened 2021-03-19
Add section on ThreadContext propagation details

cschubertcs:
## What Has Changed?

After some investigation (together with @MatKuhr) we found out that the `AuthTokenAccessor.executeWithAuthToken` really only replaces the AuthToken, but not the derived properties, like Tenant and Principal.

Therefore we deemed it necessary to extend the documentation with that information.

## Manual Checks?

### DIFF
--- a/docs/java/features/multi-tenancy/thread-context.mdx
+++ b/docs/java/features/multi-tenancy/thread-context.mdx
@@ -62,6 +62,33 @@ In that case you can override the current tenant explicitly:
 TenantAccessor.executeWithTenant(customTenant, () -> doStuff());
 ```
 
+:::caution
+Be aware, that the _executeWith_ methods shown above only replaces the given property, but does not update properties derived from it.
+
+Example: You have a special _AuthToken_, that you propagate with _AuthTokenAccessor.executeWithAuthToken_.
+Inside the given code block only the _AuthToken_ will be replaced, while e.g. the _Tenant_ is the same as in the original context.
+
+If you want to start a fresh context based on a given _AuthToken_, for example accessing information of the provider tenant while in a subscriber context, have a look at this code:
+
+```java
+
+Tenant retrieveProviderTenant()
+{
+    // retrieves an access token from the provider context
+    AuthToken providerXsuaaAccessToken = AuthTokenAccessor.getXsuaaServiceToken();
+    return new ThreadContextExecutor()
+        // don't reuse the original context
+        .withoutParentThreadContext()
+        // add the provider token into the new context
+        .withListeners(new AuthTokenThreadContextListener(providerXsuaaAccessToken))
+        // return the actual provider tenant
+        .execute(TenantAccessor::getCurrentTenant);
+}
+
+```
+
+:::
+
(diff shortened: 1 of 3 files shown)


docs/java/features/multi-tenancy/thread-context.mdx
```diff
@@ -93,3 +93,30 @@ Be cautious with long-running, asynchronous operations.
...
+:::note
+Be aware, that the _executeWith_ methods shown above only replaces the given property, but does not update properties derived from it.
+
+Example: You have a special _AuthToken_, that you propagate with _AuthTokenAccessor.executeWithAuthToken_.
+Inside the given code block only the _AuthToken_ will be replaced, while e.g. the _Tenant_ is the same as in the original context.
+
+If you want to start a fresh context based on a given _AuthToken_, for example accessing information of the provider tenant while in a subscriber context, have a look at this code:
+
+```java
+{
```
newtork, line 106: Unnecessary `{` and `}` at the end(?)

cschubertcs: Fair, fixed!

docs/java/features/multi-tenancy/thread-context.mdx
```diff
@@ -93,3 +93,30 @@ Be cautious with long-running, asynchronous operations.
 A propagated Thread context will only persist as long as the Thread lives that it was created on.
 So when the parent Thread dies the context will cease to exist and no longer be available in any of the Threads.
 :::
+
+:::note
```
MatKuhr, line 97: Note lgtm, good 2 know that you can put full code blocks into it 👍🏻

docs/java/features/multi-tenancy/thread-context.mdx
```diff
@@ -93,3 +93,30 @@ Be cautious with long-running, asynchronous operations.
 A propagated Thread context will only persist as long as the Thread lives that it was created on.
```
MatKuhr, line 93: while at it I think this here really should be a `:::caution` instead. Could you change it as well?

docs/java/features/multi-tenancy/thread-context.mdx
MatKuhr: Personally I would put this note slightly up though, at the end of the `Storing Information` section. Because that is actually what we want to achieve here where as the `asynchronous` part is more focused on retaining the information across multiple threads.

cschubertcs:
Comments incorporated!

There is now just the open question from @newtork over Slack:
> I'm not sure whether we want to promote this "workaround" when in fact we would wish for a new API like:
> ```
> TenantAccessor.getXsuaaProviderTenant()
> ```

MatKuhr:
I think this is API that should be specific to CF because getting the provider tenant is not possible nor useful in other runtimes. 

And perhaps `XsuaaProviderTenant` it is even too specific, because we might also have IAS.

Generally I don't think it is really a workaround, it's just not as convenient as it could potentially be

MatKuhr (approved):

PR #215, opened 2021-04-12
Annotate only modified lines

artemkovalyov:
## What Has Changed?

This is a pipeline update to benefit from the official Vale cation instead of a custom step to install it via brew. 
The install of Vale started to intermittently fail via brew due to unknown reasons and I found this action to be a good solution and enhance our use of it.

I tested it manually with a couple of PR and it worked fine. The annotations are added only to changed lines and visible in your "files" PR review screen. No need to switch to the Actions tab.
image


florian-richter (approved):
Seems like a sensible change. I also hijacked this to fix a typo in the warning message of the "One line per sentence" rule (not sure how I misspelled that to begin with...)

PR #236, opened 2021-04-25
Troubleshooting for the missing `execute` in some function/action imports.

jjtang1985:
## What Has Changed?

Add a troubleshooting section for explaining the missing `execute` method for some function/action imports and suggest to use `executeRaw` instead.

## Manual Checks?

- [ ] You verified all new and changed links still work (changing the `id` or name of a file can break links)
- [ ] You have updated the [feature matrix](https://github.com/SAP/cloud-sdk/blob/main/docs/components/data/features.js) if you add documentation on a new feature or otherwise required.

### DIFF
--- a/docs/js/features/odata/common/function-imports/entity-return-type.mdx
+++ b/docs/js/features/odata/common/function-imports/entity-return-type.mdx
@@ -1,5 +1,5 @@
 :::note Troubleshooting
-For some function/action-imports, the `execute()` method might be missing as intended.
+For some OData function or action imports, the `execute()` method might be missing as intended.
 Typically, this might happen when an entity type is shared by multiple entity sets and is used as the return type of function/action imports.
 You can then use `executeRaw` for getting the raw response and deserialize by your own.
 :::
--- a/docs/js/features/odata/common/function-imports/entity-return-type.mdx
+++ b/docs/js/features/odata/common/function-imports/entity-return-type.mdx
@@ -1,5 +1,5 @@
 :::note Troubleshooting
 For some OData function or action imports, the `execute()` method might be missing as intended.
-Typically, this might happen when an entity type is shared by multiple entity sets and is used as the return type of function/action imports.
+Typically, this might happen when an entity type is shared by multiple entity sets and is used as the return type of OData function or action imports.
 You can then use `executeRaw` for getting the raw response and deserialize by your own.
 :::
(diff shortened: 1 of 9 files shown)


docs/js/features/odata/common/function-imports/entity-return-type.mdx
```diff
@@ -0,0 +1,5 @@
+:::note Troubleshooting
+For some function/action-imports, the `execute()` method might be missing as intended.
```
artemkovalyov, line 2:
```suggestion
For some OData function or action imports, the `execute()` method might be missing as intended.
```

docs/js/features/odata/common/function-imports/entity-return-type.mdx
```diff
@@ -0,0 +1,5 @@
...
+Typically, this might happen when an entity type is shared by multiple entity sets and is used as the return type of function/action imports.
```
artemkovalyov, line 3:
```suggestion
Typically, this might happen when an entity type is shared by multiple entity sets and is used as the return type of OData function or action imports.
```

docs/js/features/odata/common/function-imports/entity-return-type.mdx
```diff
@@ -0,0 +1,5 @@
...
+You can then use `executeRaw` for getting the raw response and deserialize by your own.
```
artemkovalyov, line 4:
```suggestion
In such a case you can use `executeRaw` method for getting the raw response returned after invoking the OData function or action via type-safe API and deserialize it on your own.
```

artemkovalyov: I hope I guessed correctly that everywhere above you're talking about OData functions and actions? Also, the last one means calling our type-safe API for request building but using execute raw for a full response, right?

artemkovalyov (changes requested):
Mostly LTGM, @jjtang1985 , check my suggestions to see if I understood them correctly and if they are making it easier to read.

docs/js/features/odata/common/function-imports/entity-return-type.mdx
jjtang1985: Yes, your understanding is correct.
The "odata" are omitted as the documentation is under odata section/context, but adding them explicitly makes it less confusing.
I'll commit with the suggestions.

artemkovalyov (approved):
LGTM. For new customers, explicit is good regardless of the section. Thanks for the nice update.

PR #270, opened 2021-05-10
Very often SAP Cloud SDK is too verbose and short version is fine

artemkovalyov:
## What Has Changed?

I want to allow SDK as a short mentioning for the SAP Cloud SDK


artemkovalyov:
Opinions changed 50/50, we'll keep calling the SDK by Name, Surname, and Father's name.

florian-richter:
I found a way to add a `vale` rule that checks the naming center guidelines in a compliant way:

https://docs.errata.ai/vale/styles#conditional

This should allow to check that you need to mention SAP Cloud SDK at least once before using "SDK". We would need to write a new rule and adjust the existing rule, but it may be worth the effort 😊

PR #282, opened 2021-05-19
Include new feature in docs

tomfrenken:
## What Has Changed?

The new config feature is now added to the documentaiton.


artemkovalyov:
@tomfrenken, I think you have to add failing terms here: https://github.com/SAP/cloud-sdk/blob/main/styles/Vocab/SAP/accept.txt

artemkovalyov (approved):
Looks Ok

PR #332, opened 2021-06-16
We don't need this redirects anymore. Some of them are wrong

artemkovalyov:
## What Has Changed?

We had redirects to cover for early changes in the site structure. We don't need them anymore. The structure has changed even more which makes some of the redirects redundant.

I also update Docusaurus to the next version and cleaned up configs.

## Manual Checks?

~~- [] You have updated the [feature matrix](https://github.com/SAP/cloud-sdk/blob/main/docs/components/data/features.js) if you add documentation on a new feature or otherwise required.~~


docs/components/data/features.js
```diff
@@ -56,7 +56,7 @@ export const features = [
     },
     js: {
       status: ':heavy_check_mark:',
-      docsLink: `[docs](${baseUrl}/js/features/odata/use-odata-v2-type-safe-client-for-javascript-typescript)`,
+      docsLink: `[docs](${baseUrl}/js/features/odata/type-safe-client-odata-v2.mdx)`,
```
Johannes-Schneider, line 59: This does not work when I'm testing it locally.

docs/components/data/features.js
```diff
@@ -70,7 +70,7 @@ export const features = [
     },
     js: {
       status: ':heavy_check_mark:',
-      docsLink: `[docs](${baseUrl}/js/features/odata/use-odata-v4-type-safe-client-for-javascript-typescript)`,
+      docsLink: `[docs](${baseUrl}/js/features/odata/type-safe-client-odata-v4.mdx)`,
```
Johannes-Schneider, line 73: This does not work when I'm testing it locally.

docs/js/features/odata/overview.mdx
```diff
@@ -18,8 +18,8 @@ keywords:
 
 | OData protocol version                                              | SAP Cloud SDK for JavaScript |
 | ------------------------------------------------------------------- | ---------------------------- |
```
Johannes-Schneider, line 20: Could you please adjust the spacing here?

Johannes-Schneider (changes requested):
The two JS links in the feature matrix do not work.
Everything else looks good to me.

docs/components/data/features.js
artemkovalyov: WoW, thanks for the diligent review, @Johannes-Schneider

docs/js/features/odata/overview.mdx
artemkovalyov: I'll run prettier, it should fix it.

FrankEssenberger (approved):
LGTM - however, we should have a pipeline step that ensures the local startup is working right? Johannes manual test is not save enough.

artemkovalyov:
> LGTM - however, we should have a pipeline step that ensures the local startup is working right? Johannes manual test is not save enough.

@FrankEssenberger, we run a build on every PR which should catch everything for the static site. The release pipeline does deploy step after merge. If it fails new release is not deployed despite the merge. It means we can ensure the site will run.

If you're about links, they should be checked automatically. I'll check the configs because before broken links were caught automatically.

docs/components/data/features.js
artemkovalyov:
```suggestion
      docsLink: `[docs](${baseUrl}/js/features/odata/use-odata-v2-type-safe-client-for-javascript-typescript)`,
```

docs/components/data/features.js
artemkovalyov:
```suggestion
      docsLink: `[docs](${baseUrl}/js/features/odata/use-odata-v4-type-safe-client-for-javascript-typescript)`,
```

Johannes-Schneider (approved):
LGTM.

artemkovalyov:
@Johannes-Schneider, @FrankEssenberger, I found the reason the broken links were not identified. They are generated via the feature matrix and avoid regular check Docusarus runs when processing MDX. For that same reason, a slug (id) has to be used instead of the file name.

PR #353, opened 2021-06-29
Improve Feedback Snippet Visibility

Johannes-Schneider:
## What Has Changed?

This PR adds the Feedback Snippet to the Table of Contents (rhs) in a central place instead of adding it manually to individual sites.

**Hint for Reviewers**: Please concentrate your review on the following files:
* `src/css/custom.css`
* `src/sap/common-utils/DocsRating/docRating.css`
* `src/sap/common-utils/DocsRating/index.js`
* `src/theme/TOC.css`

## Manual Checks?


src/css/custom.css
```diff
@@ -71,6 +71,11 @@
   top: -3rem;
 }
 
+.toc-wrapper {
+  position: fixed;
+  top: calc(var(--ifm-navbar-height) + 1rem);
```
newtork, line 76: I wasn't aware this is possible (already)

KavithaSiva:
I checked this out locally and I liked that the feedback snippet is added to the Table of Contents component. Good Work!

One thing I noticed was, for extra long pages, theTable of contents component overrides into the footer component.
I am not sure if this is a Docusaurus bug or some setting from our side can fix this:

Screenshot 2021-06-29 at 17 32 59

src/sap/common-utils/DocsRating/index.js
```diff
@@ -21,19 +21,19 @@ const DocsRating = ({ label }) => {
...
+                    </svg>
+                    <svg
+                        className="i_positive"
+                        alt="Like"
+                        onClick={() => giveFeedback(1)}
+                        xmlns="http://www.w3.org/2000/svg" width="32px" height="32px" fill="currentColor" viewBox="0 0 32 32">
+                        <path d="M 16.6875 3 L 16.375 3.28125 L 9.59375 10 L 5 10 L 5 26 L 21.84375 26 C 23.253906 26 24.484375 25.003906 24.78125 23.625 L 26.9375 13.625 C 27.332031 11.777344 25.886719 10 24 10 L 18.25 10 L 18.4375 9.25 C 18.640625 9.09375 18.769531 9.027344 19.0625 8.625 C 19.53125 7.984375 20 6.992188 20 5.65625 C 20 4.230469 18.710938 3 17.09375 3 Z M 17.40625 5.09375 C 17.828125 5.175781 18 5.347656 18 5.65625 C 18 6.558594 17.726563 7.117188 17.46875 7.46875 C 17.210938 7.820313 17.03125 7.90625 17.03125 7.90625 L 16.6875 8.09375 L 16.5625 8.5 L 15.96875 10.75 L 15.65625 12 L 24 12 C 24.660156 12 25.105469 12.574219 24.96875 13.21875 L 22.84375 23.21875 C 22.742188 23.6875 22.320313 24 21.84375 24 L 11 24 L 11 11.40625 Z M 7 12 L 9 12 L 9 24 L 7 24 Z"/>
```
newtork, line 36: Out of interest, why is the new vector graphic "shorter"?

MatKuhr:
@KavithaSiva since also the table of contents is extending too far I'd say it's a problem with docusaurus

UI LGTM, no idea about anything CSS :D

artemkovalyov:
> @KavithaSiva since also the table of contents is extending too far I'd say it's a problem with docusaurus
> 
> UI LGTM, no idea about anything CSS :D

I think the snippet can break the layout as the menu doesn't assume smth is placed there. It can be most probably fixed but it's not the best placement IMHO.

Johannes-Schneider:
@KavithaSiva Nice catch with the overflowing table of content! That was actually my bad. To fix this, unfortunately, I needed to take over some of the original code (link in source) to get rid of the property `position: sticky` that would be set within the `OriginalTOC`.
@artemkovalyov, as I'm a total amateur with Javascript/HTML/CSS, maybe you can point out a more elegant/robust solution to remove that property?

@artemkovalyov wrote:
> it's not the best placement IMHO

Do you have a suggestion on where this snippet could be placed alternatively?

src/css/custom.css
Johannes-Schneider: This is taken from the default docusaurus theme ([here](https://github.com/facebook/docusaurus/blob/41b78466dabd1e2ee5c39672a4ee44030999eada/packages/docusaurus-theme-classic/src/theme/TOC/styles.module.css#L12)), so I would assume our docs to not work at all if that isn't supported... 
But, honestly, I actually have no clue 🙈

artemkovalyov: @newtork, CSS vars are a standard for quite some time now. Works like a charm.

artemkovalyov:
> @KavithaSiva Nice catch with the overflowing table of content! That was actually my bad. To fix this, unfortunately, I needed to take over some of the original code (link in the source) to get rid of the property `position: sticky` that would be set within the `OriginalTOC`.
> @artemkovalyov, as I'm a total amateur with Javascript/HTML/CSS, maybe you can point out a more elegant/robust solution to remove that property?
> 
> @artemkovalyov wrote:
> 
> > it's not the best placement IMHO
> 
> Do you have a suggestion on where this snippet could be placed alternatively?

@Johannes-Schneider, I'm not much of a wizard either:) Without looking you can fix classed in custome.css or have your own CSS module. I'd still do it now.

Regarding placement: top of an article and bottom of an article. If the top doesn't look well on mobile we can keep it only for desktop and hide it on mobile.

Check this Google support article for hints: https://cloud.google.com/compute/docs/quickstart-linux

artemkovalyov:
> **Unrelated to the contents of this PR.** But, I really like that in the page you linked @artemkovalyov they ask for more feedback on why users press a certain icon.

I totally agree, @Johannes-Schneider. But everything has its price. If we see a good traction we can think of this.

Johannes-Schneider:
@artemkovalyov, I reworked the placement of the snippet. It is now (very similar to what Google does) on the very top and the very bottom of each page:

localhost_3000_cloud-sdk_docs_java_features_rest_clients_scp-workflow-rest-api
localhost_3000_cloud-sdk_docs_java_features_rest_clients_scp-workflow-rest-api (1)

Unfortunately, the state between these two individual components is not shared. That means that users may vote twice for the same page, even without reloading. But, IMHO, since we do not have effective means to prevent users from reloading our docs and just keep voting over and over again, we can ignore that downside.

artemkovalyov:
> @artemkovalyov, I reworked the placement of the snippet. It is now (very similar to what Google does) on the very top and the very bottom of each page:
> 
> localhost_3000_cloud-sdk_docs_java_features_rest_clients_scp-workflow-rest-api
> localhost_3000_cloud-sdk_docs_java_features_rest_clients_scp-workflow-rest-api (1)
> 
> Unfortunately, the state between these two individual components is not shared. That means that users may vote twice for the same page, even without reloading. But, IMHO, since we do not have effective means to prevent users from reloading our docs and just keep voting over and over again, we can ignore that downside.

@Johannes-Schneider, here's how you can sync them using React Hooks: https://codesandbox.io/s/great-newton-y36cu
I don't insist on having that. It's just FYI.

artemkovalyov (approved):
LGTM, here's an example how to sync two components if you'd like to give it a try: https://codesandbox.io/s/great-newton-y36cu

PR #389, opened 2021-07-26
Fix API docs links to the new specs

artemkovalyov:
## What Has Changed?

Fix API Docs links


jjtang1985 (approved):
LGTM, if this happens frequently, we should consider some tests for broken links.

PR #668, opened 2022-02-02
chore: run prettier and linter

FrankEssenberger:
## What Has Changed?

I realized the the prettier and linter have not been executed. I think we should put this in the pipeline to enforce this.


artemkovalyov (approved):
Yep, the best would be having GitHooks to avoid running it on CI actually.

PR #670, opened 2022-02-03
Fix workaround

newtork:
## What Has Changed?

I think providing the headers explicitly does not bring any benefit or does it?

Why do we read `RequestHeaderContainer` from parent thread-context and pass it manually in the child thread-context - assuming inheritance would've done that anyway for us.


Johannes-Schneider:
> Why do we read RequestHeaderContainer from parent thread-context and pass it manually in the child thread-context - assuming inheritance would've done that anyway for us.

This workaround is also meant to enable scenarios where the "inheritance" does not work (i.e. the "child" task is started **after** the parent thread has already been cleared).
Therefore, accessing and storing the reference to the `RequestHeaderContainer` while the "parent" Thread is still alive is mandatory.

MatKuhr:
Closing this due to the limitation of this approach

PR #675, opened 2022-02-07
Java release policy

artemkovalyov:
## What Has Changed?

Adding public release policy for SDK Java

## Manual Checks?


docs/java/release-policy.mdx
```diff
@@ -0,0 +1,142 @@
...
+|:--------|-------------|:----------------------|:--------------|---------------|
+| 3.x     | Released GA | January 2019 (fix me) | N/A           |               |
+| 4.x     | Planned     | July 2022             | TBD           | TBD           |
+| 5.x     | Planned     | January 2023          | TBD           | TBD           |
+| 6.x     | Planned     | July 2023             | TBD           | TBD           |
+
+
+## Minor and Major Release Policy
+
+We follows [Semantic Versioning](https://semver.org/) for our releases.
```
KavithaSiva, line 43:
```suggestion
We follow [Semantic Versioning](https://semver.org/) for our releases.
```

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,142 @@
...
+We sometimes change our release frequency to:
+
+  - Provide a **hot fix** release.
+    It is delivered as a patch version to the latest minor release when a critical bug fix is necessary.
+  - Accommodate a longer development cycle.
+    We can extend an interval between **minor version** releases.
+
+:::tip  "Upgrading between minor releases"
+
+Usually, an upgrade between minor releases do not involve any effort.
```
KavithaSiva, line 59:
```suggestion
Usually, an upgrade between minor releases does not involve any effort.
```

docs/java/release-policy.mdx
KavithaSiva, line 43:
```suggestion
We follow [Semantic Versioning](https://semver.org/) for our releases.
```

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,142 @@
...
+We follow [Semantic Versioning](https://semver.org/) for our releases.
+
+### Minor release schedule
+
+Our regular **minor version** release happens every **two weeks**.
+It includes new functionality, fixes, and maintenance updates that we communicate in the [release notes](./release-notes.mdx).
+
+We sometimes change our release frequency to:
+
+- Provide a **hot fix** release.
```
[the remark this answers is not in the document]
MatKuhr: we rarely release a patch version in the regular bi-weekly cycle

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,142 @@
...
+
+We intend to release a new **major version** every **six months** to introduce significant feature upgrades, refactor existing features, and clean up deprecated code.
+
+- **Major version** release interval is indicative and can be changed.
+- We will announce planned date and scope of the next **major version** release in this [document](#planned-major-release-schedule) and [release notes](./release-notes.mdx) at least 3 months before the planned release date.
+
+:::caution "Upgrading between major versions"
+
+Major releases include incompatible API changes.
+An upgrade will usually involve a certain effort to make your code benefit from new APIs.
```
MatKuhr, line 74: Not sure if I'd say that. There may be many customers who don't need to change anything with a particular upgrade. E.g. when removing CXSDK. Maybe we can rephrase a little, pointing out that not every upgrade will require adaptations & for each upgrade we will have detailed instructions where changes are needed.

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,142 @@
...
+:::tip
+
+We will always provide a branch with the latest snapshot of the documentation before the major version release.
+You can use that branch to host documentation for the earlier version locally.
+
+:::
+
+### Fixes, Support, and Maintenance
+
+We provide new features and fixes only for the **latest major release** of the SAP Cloud SDK for Java.
```
MatKuhr, line 124: I feel that one sentence is a bit short for this topic. I'd elaborate a bit more. E:g. what are "fixes"? Also, as a customer I'd not be happy to hear that if I am on the latest version e.g. 3.70  and 4.0 is released I immediately loose "support" for 3.70 until I upgrade.

MatKuhr (commented):
We could also add a note for internal customers about the Beta Phase. 

Otherwise I'd restructure slightly & expand a few details. Otherwise I think we can publish this

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,142 @@
...
+| Version | Status      | Release Date          | Planned scope | Upgrade guide |
+| :------ | ----------- | :-------------------- | :------------ | ------------- |
+| 3.x     | Released GA | January 2019 (fix me) | N/A           |               |
+| 4.x     | Planned     | July 2022             | TBD           | TBD           |
+| 5.x     | Planned     | January 2023          | TBD           | TBD           |
+| 6.x     | Planned     | July 2023             | TBD           | TBD           |
+
+## Minor and Major Release Policy
+
+We follow [Semantic Versioning](https://semver.org/) for our releases.
```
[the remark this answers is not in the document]
artemkovalyov: Didn't get what kind of top?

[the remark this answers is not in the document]
artemkovalyov: I'd keep the table with the release plan on top at all times. That will be the only reason for users to return to this page and at the end that's the most relevant information.

docs/java/release-policy.mdx
KavithaSiva: I agree with Matthias here, loosing support immediately for the previous major version when a new one is released could be an issue for customers.
What about the idea of extending support for a fixed time say for e.g 2-3 months, but maybe only fix critical security issues for the previous version then?

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,141 @@
...
+It includes new functionality, fixes, and maintenance updates that we communicate in the [release notes](./release-notes.mdx).
+
+We sometimes change our release frequency to:
+
+- Provide a **hot fix** release.
+  It is delivered as a patch version to the latest minor release when a critical bug fix is necessary.
+- Accommodate a longer development cycle.
+  We can extend an interval between **minor version** releases.
+
+:::tip "Upgrading between minor releases"
```
Johannes-Schneider, line 55: Is there a specific reason to use quotes here?

```suggestion
:::tip Upgrading between minor releases
```

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,141 @@
...
+:::
+
+### Major release schedule
+
+We intend to release a new **major version** every **six months** to introduce significant feature upgrades, refactor existing features, and clean up deprecated code.
+
+- **Major version** release interval is indicative and can be changed.
+- We will announce planned date and scope of the next **major version** release in this [document](#planned-major-release-schedule) and [release notes](./release-notes.mdx) at least 3 months before the planned release date.
+
+:::caution "Upgrading between major versions"
```
Johannes-Schneider, line 70:
```suggestion
:::caution Upgrading between major versions
```

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+| Version | Status      | Release Date          | Planned scope | Upgrade guide |
+| :------ | ----------- | :-------------------- | :------------ | ------------- |
+| 3.x     | Released GA | August 2019 | N/A           |               |
+| 4.x     | Planned     | July 2022             | TBD           | TBD           |
+| 5.x     | Planned     | January 2023          | TBD           | TBD           |
+
+## Minor and Major Release Policy
+
+We follow [Semantic Versioning](https://semver.org/) for our releases.
```
Johannes-Schneider, line 41: This section does not provide any new information and should either be removed entirely (we already outlined that we are using semantic versioning) or merged together with the following sections IMO.

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,141 @@
...
+All the changes are incremental and backward compatible.
+We do not remove any deprecated code before the next major release.
+
+:::
+
+### Major release schedule
+
+We intend to release a new **major version** every **six months** to introduce significant feature upgrades, refactor existing features, and clean up deprecated code.
+
+- **Major version** release interval is indicative and can be changed.
```
Johannes-Schneider, line 67: [Minor]

Personally, I'd prefer the following wording:

```suggestion
- **Major version** release interval is indicative and may change.
```

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,141 @@
...
+- We will announce planned date and scope of the next **major version** release in this [document](#planned-major-release-schedule) and [release notes](./release-notes.mdx) at least 3 months before the planned release date.
```
Johannes-Schneider, line 68: > at least 3 months before the planned release date

Not sure if that is actually realistic to do (maybe depends on what we understand as "release notes" here).
What about something like _(feature) scope_ instead of _"release notes"_?

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+:::caution "Upgrading between major versions"
+
+Major releases include incompatible API changes.
```
Johannes-Schneider, line 72: This is not a given, right?

```suggestion
Major releases may include incompatible API changes.
```

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,141 @@
...
+### Documentation
+
+The documentation is always up to date with the **latest major release**.
+We provide documentation for the two most recent major versions at the same time.
+We will use tabs to demonstrate a difference between them where required.
+See example below
+
+<Tabs groupId="version" defaultValue="major" values={[ { label: 'SDK 4.x', value: 'major' }, { label: 'SDK 3.x', value: 'deprecated' } ]}>
+
+<TabItem value="major">
```
Johannes-Schneider, line 90: I have some doubts with regards to whether this is actually a good and maintainable approach.
Especially in cases where concepts change in between major versions (e.g. our ThreadContext concept might get a bigger refactoring in the future) this approach could lead to an entire page being wrapped inside such a table.
Even more, what if our current documentation structure does no longer fit the actual code? We then had to somehow introduce some ugly workarounds and/or have semi-optimal documentation.

Why are we not using the versioning feature of Docusaurus?

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+Deprecated version documentation
+
+```
+
+</TabItem>
+
+</Tabs>
+
+With every new major version release, the oldest documented version is removed from the documentation portal.
```
Johannes-Schneider, line 112: IMO this is a big disadvantage of the suggested approach.
Not only do we have to (potentially) add a whole lot of documentation for new/updated functionality, we would also need to remove all of the older documentation (which is still valid and might help customers who are still on an older version).

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,141 @@
...
+We will always provide a branch with the latest snapshot of the documentation before the major version release.
+You can use that branch to host documentation for the earlier version locally.
+
+:::
+
+### Fixes, Support, and Maintenance
+
+We provide new features and fixes only for the **latest major release** of the SAP Cloud SDK for Java.
+
+:::tip "Upgrade early"
```
Johannes-Schneider, line 125:
```suggestion
:::tip Upgrade early
```

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,141 @@
...
+- protection of your development process from disruptions in the future.
+  Upgrading gets more complex if you have to migrate between more than one major version.
+
+:::
+
+## Support and feedback
+
+Let us know if you have any questions about our release policy.
+We usually respond within one day.
+Check our [support channels](../overview/support.mdx) and chose the one working best for you.
```
Johannes-Schneider, line 141:
```suggestion
Check our [support channels](../overview/support.mdx) and choose the one working best for you.
```

docs/java/release-policy.mdx
MatKuhr: +1, also we don't have to document this here at this point in time. This only becomes relevant once we start releasing documentation.

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+</TabItem>
+
+</Tabs>
+
+With every new major version release, the oldest documented version is removed from the documentation portal.
+
+:::tip
+
+We will always provide a branch with the latest snapshot of the documentation before the major version release.
```
MatKuhr, line 116: pretty strong statement, I would not make that at this point in time

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+Our regular **minor version** release happens every **two weeks**.
+It includes new functionality, fixes, and maintenance updates that we communicate in the [release notes](./release-notes.mdx).
+
+We sometimes change our release frequency to:
+
+- Provide a **hot fix** release.
+  It is delivered as a patch version to the latest minor release when a critical bug fix is necessary.
+- Accommodate a longer development cycle.
+  We can extend an interval between **minor version** releases.
```
artemkovalyov, line 53:
```suggestion
- Provide a patch version to the latest minor release when a critical bug fix is necessary.
- Accommodate a longer development cycle.
  We can extend an interval between **minor version** releases.
```

docs/java/release-policy.mdx
artemkovalyov: We discussed that we can simply put a release TAG to the docs and then explain how to fetch it and run npm install, npm run start. This doesn't seem like a challenge. Or did you mean smth different?

docs/java/release-policy.mdx
artemkovalyov: That's why we provide a release TAG in our Docs so that customers can self-host the docs for their version.
Public docs will have only 2 versions at all times - prev and current.

artemkovalyov: Most of the docs will be relevant and won't change. Only parts that are touched by the change will be affected.

docs/java/release-policy.mdx
artemkovalyov: That's a part of the strategy and I see it relevant. On the other hand, if you see it differently suggest options, and let's discuss. This worked well for JS, we had a bigger discussion there.

Where concepts do not meet we can have a case-by-case approach. It might be not as complex as you think it is.

docs/java/release-policy.mdx
artemkovalyov: It's a given. That is what defines major release. The fact they might not touch you code line doesn't deny the fact.

docs/java/release-policy.mdx
artemkovalyov: This will be a link to the separate document, not added inline to the policy or the release notes. Is that your concern?

docs/java/release-policy.mdx
artemkovalyov: I think both versions work, but the quote less looks cleaner.

docs/java/release-policy.mdx
artemkovalyov: I'd put it in the separate doc and only after we find it useful. This feels like overplanning to me. We can fix/improve it at any time it becomes obvious as a value for the customer. Now I doubt it.

I also wouldn't bind us with the last line in your suggestion. We most probably don't do it. "For a while" - is not a timeframe.

docs/java/release-policy.mdx
Johannes-Schneider: No, my concern is more about the _"3 months before the planned release date"_.
I'm worried that we won't have concrete _"release notes"_ at that point in time.
Maybe it's just the phrase _"release notes"_, which for me means very detailed descriptions of what will be changed (similar to what we already do for our regular releases).

Johannes-Schneider: Ahh.... now I see. 😅
I misunderstood your sentence as follows:

"We are going to announce the scope **and** release notes 3 months ahead of time"

while you are actually saying

"We are going to announce the scope on this page (here) and on the release notes page".


Sorry for the confusion!

Johannes-Schneider (approved):
LGTM but lets get at least a second opinion before merging.

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,142 @@
...
+  - strategy
+  - major
+  - release
+  - support
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+:::info relevance
```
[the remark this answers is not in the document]
MatKuhr: Hmm, this is explicitly under the Java section, so why should it be here and not on every other article...

Personally I don't mind it being here, except that it feels a little awkward to be at the very top, but I guess there is not really a better place to put it. So I'd be just fine to leave it here...

docs/java/release-policy.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+We will always provide a branch with the latest snapshot of the documentation before the major version release.
+You can use that branch to host documentation for the earlier version locally.
+
+:::
+
+### Maintenance and Support
+
+- We provide new features and fixes **only** for the **latest major release** of the SAP Cloud SDK for Java.
+- We continue to [support](./../overview/support.mdx) customers using any version of the SAP Cloud SDK for Java.
```
MatKuhr, line 122: to me this reads much clearer now, thanks!

MatKuhr (approved):

PR #684, opened 2022-02-11
Improve Logging Recommendations for Java

MatKuhr:


Johannes-Schneider (approved):
LGTM.

docs/java/guides/logging-overview.mdx
```diff
@@ -234,12 +234,31 @@ In the case that you want to log all requests sent and responses received by the
...
+
+```bash
+{ROOT: INFO, com.sap.cloud.sdk: INFO, org.apache.http: DEBUG, org.apache.http.wire: ERROR}
+```
+
+You can set the `SET_LOGGING_LEVEL` environment variable via your deployment descriptor or via the CF CLI.
+
+:::tip Dynamically Changing Log Levels at Runtime
+Using the `SET_LOGGING_LEVEL` requires at least a restart of your application for changes to take effect.
+Alternatively, you can also change the log levels dynamically and even on a per-request basis via [Dynamic Logging](https://github.com/SAP/cf-java-logging-support/wiki/Dynamic-Log-Levels).
```
Johannes-Schneider, line 254: Should we verify that ourselves first before recommending it to customers?

MatKuhr: No, we are just linking here for the interested reader and are not claiming anything specific. I'd say this is advanced stuff and the page linked is well maintained it seems, so I'd be confident.

MatKuhr: I mean I would be happy to try it out, but not see it as mandatory :D

PR #681, opened 2022-02-08
doc: Add documentation on transpilation

marikaner:
Closes SAP/cloud-sdk-backlog/issues/577
## What Has Changed?

I added some docs on how the OData generator is influenced by dependencies and transpilation. I am not sure whether it is explained too complicated...
After the first feedback round I would add a similar documentation on OpenAPI.

## Manual Checks?

- [ ] Text adheres to the style guide (`vale docs/`)
  - Every sentence is on its own line
  - Headings use [title capitalization](https://capitalizemytitle.com/style/AP/#) (applies also to the sidebar and title of a document)
  - You followed [naming center](https://www.sapbrandtools.com/naming-center/#/dashboard) guidelines when referring to SAP products (e.g. SAP S/4HANA)
- [ ] You checked your spelling and grammar (consider using Grammarly, see `CONTRIBUTING.md`)
- [ ] You formatted all changed files with prettier (`npm run prettier`)
- [ ] You tested if the documentation still builds (`npm run build`)
- [ ] You verified all new and changed links still work (changing the `id` or name of a file can break links)
- [ ] You have updated the [feature matrix](https://github.com/SAP/cloud-sdk/blob/main/docs/components/data/features.js) if you add documentation on a new feature or otherwise required.

PR #694, opened 2022-02-16
docs: Add documentation on transpilation to openapi generator

marikaner:
## What Has Changed?

Same documentation as in #681, but for OpenAPI.
I adjusted it to be semantically correct, but structure and content is the same.


florian-richter (approved):
Maybe put the `npm install -D @sap-cloud-sdk/openapi` into a info box, otherwise LGTM

PR #701, opened 2022-02-21
[Java] RequestHeaderAccessor/-Container Documentation

Johannes-Schneider:
## What Has Changed?

This PR adds detailed documentation about our "new" `RequestHeaderAccessor` and the `RequestHeaderContainer` API.

## Manual Checks?


docs/java/features/connectivity/header-container.mdx
```diff
@@ -0,0 +1,133 @@
...
+  - sap
+  - cloud
+  - sdk
+  - headers
+  - java
+  - connectivity
+---
+
+The SAP Cloud SDK for Java offers a broad spectrum of convenience _accessors_.
+These _accessors_ can be used to get information such as the currently active _tenant_, the _principal_, and the _authentication token_.
```
MatKuhr, line 18: This overlaps somewhat with [this explanation](https://sap.github.io/cloud-sdk/docs/java/features/multi-tenancy/multi-tenancy-thread-context#accessing-information)

docs/java/features/connectivity/header-container.mdx
```diff
@@ -0,0 +1,133 @@
...
+Under the hood, the _accessors_ extract the requested information from the _incoming_ (user) request that is currently being processed.
+Before version `3.61.0`, the implementations were relying on Java's `Servlet` API.
```
MatKuhr, line 20: Generally speaking we only document the current state in our docs, so I'd omit this here because it's not really important for developers to know (or is it?)

docs/java/features/connectivity/header-container.mdx
```diff
@@ -0,0 +1,133 @@
...
+Unfortunately, that approach has proven to be inflexible, especially in asynchronous applications.
+Therefore, the `RequestHeaderAccessor` was introduced with version `3.61.0` of the SAP Cloud SDK.
+
+## Reading HTTP Headers
+
+The `RequestHeaderAccessor` provides convenient access to the HTTP headers that are contained in the currently processed _incoming_ (user) request:
```
MatKuhr, line 26:
```suggestion
The `RequestHeaderAccessor` provides convenient access to the HTTP headers that are contained in the currently processed _incoming_ request:
```

A request is not necessarily triggered by a user, it could just as well be a service-2-service call, e.g. a subscription call-back. If we want to clarify we could say something like:

> the HTTP request to the application that triggered the current processing, if any

docs/java/features/connectivity/header-container.mdx
```diff
@@ -0,0 +1,133 @@
...
+```
+
+The `RequestHeaderContainer` is an **immutable** collection of HTTP headers.
+Each HTTP header consists of a name and arbitrary many values.
+To access the values for a specific header, use `RequestHeaderContainer#getHeaderValues(String headerName)`.
+
+:::tip Names are Case-Insensitive
+
+The `RequestHeaderContainer` guarantees that header names are treated case-insensitively when accessing header values.
+That way, you can be assured that both calls `myRequestHeaders.getHeaderValues("EXAMPLE")` and `myRequestHeaders.getHeaderValues("example")` will yield the same values.
```
MatKuhr, line 43: This will change in the future with going to single-value headers only, right?

docs/java/features/connectivity/header-container.mdx
```diff
@@ -0,0 +1,133 @@
...
+
+:::
+
+You can also check whether a certain header is present by using `RequestHeaderAccessor#containsHeader(String headerName)` - also case-insensitive.
+
+### Usage in Other Accessors
+
+As mentioned in the beginning, prior to version `3.61.0`, the SAP Cloud SDK's convenience _accessors_, such as the `AuthTokenAccessor`, used to rely on the incoming `ServletRequest`.
```
MatKuhr, line 51: This doesn't really explain what happens if you specify headers manually & what takes precedence. Also, this actually depends on the implementation of the facades. E.g. doesn't work with CAP.

docs/java/features/connectivity/header-container.mdx
```diff
@@ -0,0 +1,133 @@
...
+        .toBuilder()
+        .withHeader("my-new-header", "value")
+        .withoutHeader("header-to-remove")
+        .replaceHeader("existing-header", "new value 1", "new value 2")
+        .build();
+```
+
+You will end up with a **new instance** of the `RequestHeaderContainer` interface that contains the updated header values.
+
+### Forwarding Headers to other Features
```
MatKuhr, line 109: "Forwarding" sounds somewhat strange here, it may give the impression that this is about forwarding headers in outgoing calls

MatKuhr (commented):
I think this article should be merged into / integrated with the general article about the [SDK context](https://sap.github.io/cloud-sdk/docs/java/features/multi-tenancy/multi-tenancy-thread-context#storing-information).

Also we might want to hold of merging this until we change the header behaviour

docs/java/features/connectivity/header-container.mdx
Johannes-Schneider: Why would we switch to single-value headers?
Are you referring to the planned change of no longer splitting header values by `,` (or cookies by `;`)?
If so, then we would still need to support multiple values because a request can still contain the same header key multiple times, right?

Johannes-Schneider:
Converted this to a draft PR since we are planning to do some API (and potentially) behavioral breaking changes.

MatKuhr (approved):
LGTM. I think we can publish this as is, even though the executeWith functionality currently doesn't work with CAP. But that is to be clarified in a different PR I'd say

PR #772, opened 2022-04-13
docs(js): add path templating doc to generate-openapi-client.mdx

ZhongpinWang:
## What Has Changed?

Ref: SAP/cloud-sdk-js#2358
Closes: SAP/cloud-sdk-backlog#52

## Manual Checks?

### DIFF
--- a/docs/js/features/openapi/generate-openapi-client.mdx
+++ b/docs/js/features/openapi/generate-openapi-client.mdx
@@ -162,6 +162,26 @@ Example:
 | `myFunction`            | `myFunction`             |
 | `other_function`        | `otherFunction`          |
 
+#### Path Templating
+Path templating can be used as:
+
+```yaml
+paths:
+  /entities/{pathParam}:
+    parameters:
+      - in: path
+        name: pathParam
+        required: true
+        type: string
+```
+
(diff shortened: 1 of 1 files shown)


marikaner (changes requested):
Sorry for coming up with this so late, I am wondering whether this documentation is really necessary. It basically states, that incorrect path parameters as of the OpenAPI specification are not supported and if parameters are missing the generation will fail.

Feel free to disagree, but I think we should not add this to our documentation, because it adds too many details, that are not relevant enough. @jjtang1985, @frank-essenberger, @florian-richter, @deekshas8, @tomfrenken opinions?

florian-richter:
Agreed, I don't think we need to inform users about this. I think this may be a case of "it just works" and you don't need to worry about it. Crazy edge cases (like `{para/m}`) can be handled in answers to support tickets.

@marikaner Good that you caught it now, even though we should have realized this before Zhongpin spent time on this...

FrankEssenberger:
For me it would be fine to close it  since it states: For correct files it works. If we keep it we should add a bit more like `For valid spec files you should not have to do something. But if you face errors of the kind XYZ this is related to the follwing requirements`.

marikaner:
Thank you for your feedback, I will go ahead and close this then.

PR #825, opened 2022-05-23
[js] add documentation on dest properties

FrankEssenberger:
## What Has Changed?

Add a short section on the destination properties which are considered by the SDK.

## Manual Checks?

~~- [ ] You have updated the [feature matrix](https://github.com/SAP/cloud-sdk/blob/main/docs/components/data/features.js) if you add documentation on a new feature or otherwise required.~~


docs/js/features/connectivity/destination.mdx
```diff
@@ -391,6 +391,47 @@ In the section above we listed already a few of the options but this sections gi
...
 
+## Destination Properties
+
+The destination object may contain additional properties.
+The properties change the behavior how the SAP Cloud SDK handles the HTTP request at runtime.
+
+### SAP Client
+
+The property `sap-client` is considered by the SAP Cloud SDK.
+This property leads to a header field `sap-client` with the specified value in the HTTP request.
```
marikaner, line 402: [pp] "This property leads to" sounds a bit strange - The SDK does something with the property, which is just there.

```suggestion
The property `sap-client` is considered by the SAP Cloud SDK.
When this property is set, it is used as the header parameter `sap-client` with the specified value in the HTTP request to the target system.
```

docs/js/features/connectivity/destination.mdx
```diff
@@ -391,6 +391,47 @@ In the section above we listed already a few of the options but this sections gi
...
+
+<img
+  alt="SAP client property on destination"
+  src={useBaseUrl('img/destination-property-sap-client.png')}
+  class="center"
+/>
+
+### Trust Configuration
+
+There are two destination property used to establish trust to systems using certificates not issued by trusted certificate authorities:
```
marikaner, line 412: [req] Typo
[req] The sentence is hard to read

```suggestion
By default, SAP BTP only trusts certain certificate authorities.
If you want to make HTTPS requests against systems that use certificates from other certificate authorities, you can configure the following properties:
```

docs/js/features/connectivity/destination.mdx
```diff
@@ -391,6 +391,47 @@ In the section above we listed already a few of the options but this sections gi
...
+
+For additional information on trust configuration have a look at the more [detailed guide](../../guides/trust-and-key-stores.mdx).
+
+<img
+  alt="Truststore location property on destination"
+  src={useBaseUrl('img/destination-property-truststore.png')}
+  class="center"
+/>
+
+<br />
```
marikaner, line 425: [q] I think to keep the theme consistent everywhere, we shouldn't do this. How bad is it?
```suggestion
```

docs/js/features/connectivity/destination.mdx
```diff
@@ -391,6 +391,47 @@ In the section above we listed already a few of the options but this sections gi
...
+### SAP Client
+
+The property `sap-client` is considered by the SAP Cloud SDK.
+This property leads to a header field `sap-client` with the specified value in the HTTP request.
+
+<img
+  alt="SAP client property on destination"
+  src={useBaseUrl('img/destination-property-sap-client.png')}
+  class="center"
+/>
```
marikaner, line 408: [q] General: I am not sure if it is worth creating images for all of these. Do you think it would be enough to just have one for all of them in the introduction of this section?

marikaner (changes requested):
Mostly minor stuff, just one sentence, that I would really try to simplify and a few questions.

docs/js/features/connectivity/destination.mdx
FrankEssenberger: Then there is no line break between the pictures.

docs/js/features/connectivity/destination.mdx
FrankEssenberger: I think so too. I will take one image and remove the other

FrankEssenberger: ALso removed the other trust pic because it was to much.

marikaner: 👍

marikaner (approved):
👍

PR #852, opened 2022-06-13
[ThreadContext] Update Async Execution Docs

MatKuhr:
This PR documents the new APIs for async executions. Furthermore, this PR removes the thread lifecycle restrictions.

### DIFF
--- a/docs/java/features/multi-tenancy/thread-context.mdx
+++ b/docs/java/features/multi-tenancy/thread-context.mdx
@@ -59,7 +59,7 @@ Please note that the `RequestScopedHttpClientCache` which is the default being u
 ## Running Asynchronous Operations
 
 As the name suggests, the `ThreadContext` is bound to a thread, more specifically to the one it was created in.
-If asynchronous operations need to access the information, it has to be propagated to the new threads.
+In consequence, other threads, i.e. asynchronous operations, cannot access the stored information if it hasn't been explicitly propagated.
 
 The SAP Cloud SDK offers a dedicated API to execute asynchronous operations while preserving the `ThreadContext`.
 
--- a/docs/java/features/multi-tenancy/thread-context.mdx
+++ b/docs/java/features/multi-tenancy/thread-context.mdx
@@ -103,7 +103,7 @@ public class AsynchronousConfiguration implements AsyncConfigurer {
 
   @Override
   public Executor getAsyncExecutor() {
-    return ThreadContextExecutors.getDefault();
+    return ThreadContextExecutors.getExecutor();
   }
 }
 
--- a/docs/java/features/multi-tenancy/thread-context.mdx
+++ b/docs/java/features/multi-tenancy/thread-context.mdx
@@ -152,8 +152,13 @@ To avoid multiple wrappings, in particular when changing multiple values, you ca
 
 ```java
 Tenant myTenant = new DefaultTenant("foo");
+Principal myPrincipal = new DefaultPrincipal("bar", Collections.emptySet(), Collections.emptyMap());
+        
 ThreadContextExecutor customExecutor = new ThreadContextExecutor()
-            .withListeners(new TenantThreadContextListener(myTenant));
+            .withListeners(
+                new TenantThreadContextListener(myTenant),
+                new PrincipalThreadContextListener(myPrincipal)
+            );
 
 ThreadContextExecutors.submit(myTask, customExecutor);
 ```
--- a/docs/java/features/multi-tenancy/thread-context.mdx
+++ b/docs/java/features/multi-tenancy/thread-context.mdx
@@ -85,7 +85,7 @@ Refer to the [dedicated section below](#spring-integration).
 :::
 
 :::caution CAP Integration
-In case you are using CAP the CAP Request Context also needs to be passed on (in addition to the `ThreadContext`).
+In case you are using CAP, note that the CAP Request Context also needs to be passed on (in addition to the `ThreadContext`).
 Please refer to [this guide](../../guides/cap-sdk-integration.mdx) on how to achieve this.
 :::
 
--- a/docs/java/features/multi-tenancy/thread-context.mdx
+++ b/docs/java/features/multi-tenancy/thread-context.mdx
@@ -105,6 +105,7 @@ public class AsynchronousConfiguration implements AsyncConfigurer {
     return ThreadContextExecutors.getDefault();
   }
 }
+
 ```
 
 You can read more about the `@Async` functionality [here](https://www.baeldung.com/spring-async).
--- a/docs/java/features/multi-tenancy/thread-context.mdx
+++ b/docs/java/features/multi-tenancy/thread-context.mdx
@@ -71,7 +71,7 @@ Callable myTask = () -> doStuff();
 Runnable myTask = () -> doStuff();
 ```
 
-The operation can be submitted via:
+The operation can be executed asynchronously via:
 
 ```java
 Future runningTask = ThreadContextExecutors.submit(myTask);
--- a/docs/java/features/multi-tenancy/thread-context.mdx
+++ b/docs/java/features/multi-tenancy/thread-context.mdx
@@ -113,7 +113,7 @@ You can read more about the `@Async` functionality [here](https://www.baeldung.c
 
 ### Configuring the Executor
 
-The `ThreadContextExecutors` class leverages an `ExecutorService` instance that can be configured.
+The `ThreadContextExecutors` class leverages a single `ThreadContextExecutorService` instance that can be configured.
 
 You can create a custom `ThreadContextExecutorService`, for example to use a different thread pool, via:
 
--- a/docs/java/features/multi-tenancy/thread-context.mdx
+++ b/docs/java/features/multi-tenancy/thread-context.mdx
@@ -56,6 +56,94 @@ That also means that the `ThreadContext` will not be used when accessing this in
 Please note that the `RequestScopedHttpClientCache` which is the default being used by the SAP Cloud SDK cannot work with this approach.
 :::
 
+## Running Asynchronous Operations
+
+As the name suggests the `ThreadContext` is bound to a thread, more specifically to the one it was created.
+If asynchronous operations need to access the information, it has to be propagated to the new threads.
+
+The SAP Cloud SDK offers a dedicated API to execute asynchronous operations while preserving the `ThreadContext`.
+
+Given any operation:
+```java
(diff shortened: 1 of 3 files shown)


docs/java/features/multi-tenancy/thread-context.mdx
```diff
@@ -56,6 +56,94 @@ That also means that the `ThreadContext` will not be used when accessing this in
...
+
+To run something in a completely new, empty context, use:
+```java
+ThreadContextExecutors.submit(myTask, new DefaultThreadContext());
+```
+
+By contrast, to pass on the current context, but modifying the tenant (and only the tenant), use:
+```java
+Tenant myTenant = new DefaultTenant("foo");
+ThreadContextExecutor customExecutor = new ThreadContextExecutor()
```
MatKuhr, line 140: TBD: Do we want to advertise this style? 

Alternatives would be:
```java
 Tenant myTenant = new DefaultTenant("foo");

Callable myTaskWithTenant = () -> TenantAccessor.executeWithTenant(myTask, myTenant);
ThreadContextExecutors.submit(myTaskWithTenant);
```
or:
```java
ThreadContext contextWithTenant = ThreadContextAccessor.getCurrentContext().duplicate().removeProperty("Tenant");
contextWithTenant.setPropertyIfAbsent("Tenant", Try.success(Property.of(myTenant)));

ThreadContextExecutors.submit(myTaskWithTenant, contextWithTenant);
```

docs/java/features/multi-tenancy/thread-context.mdx
```diff
@@ -56,6 +56,98 @@ That also means that the `ThreadContext` will not be used when accessing this in
 Please note that the `RequestScopedHttpClientCache` which is the default being used by the SAP Cloud SDK cannot work with this approach.
 :::
 
+## Running Asynchronous Operations
+
+As the name suggests the `ThreadContext` is bound to a thread, more specifically to the one it was created.
```
Johannes-Schneider, line 61:
```suggestion
As the name suggests, the `ThreadContext` is bound to a thread, more specifically to the one it was created in.
```

docs/java/features/multi-tenancy/thread-context.mdx
```diff
@@ -56,6 +56,98 @@ That also means that the `ThreadContext` will not be used when accessing this in
...
+If asynchronous operations need to access the information, it has to be propagated to the new threads.
```
Johannes-Schneider, line 62:
```suggestion
In consequence, other threads, i.e. asynchronous operations, cannot access the stored information if it hasn't been explicitly propagated.
```

docs/java/features/multi-tenancy/thread-context.mdx
```diff
@@ -56,6 +56,98 @@ That also means that the `ThreadContext` will not be used when accessing this in
...
+
+Given any operation:
+
+```java
+Callable myTask = () -> doStuff();
+// or
+Runnable myTask = () -> doStuff();
+```
+
+The operation can be submitted via:
```
Johannes-Schneider, line 74:
```suggestion
The operation can be executed asynchronously via:
```

docs/java/features/multi-tenancy/thread-context.mdx
```diff
@@ -56,6 +56,98 @@ That also means that the `ThreadContext` will not be used when accessing this in
...
+
+```java
+Future runningTask = ThreadContextExecutors.submit(myTask);
+```
+
+This will use a default `ThreadContextExecutorService` based on a `CachedThreadPool` to create a new Thread and run the task.
```
Johannes-Schneider, line 80:
```suggestion
By default, this will use a `ThreadContextExecutorService` based on a `CachedThreadPool` to create a new Thread and run the task.
```

Johannes-Schneider (commented):
Generally, the documentation makes sense to me.
Some of the formulations could be improved though.

docs/java/features/multi-tenancy/thread-context.mdx
MatKuhr: +1 on the third, I mentioned it mostly for completeness :D 

Additional downside of (1) is that the customer needs to interact with `ThreadContextExecutor` which is relatively easy to misuse (potentially less so if we apply [this improvement](https://github.wdf.sap.corp/MA/sdk/pull/7033#discussion_r4315070)).

We could also just describe both..

docs/java/features/multi-tenancy/thread-context.mdx
MatKuhr: I would keep this as is,  to emphasise that this will always use "a default executor", namely the one set via `setDefault`

docs/java/features/multi-tenancy/thread-context.mdx
```diff
@@ -56,6 +56,98 @@ That also means that the `ThreadContext` will not be used when accessing this in
...
+  public Executor getAsyncExecutor() {
+    return ThreadContextExecutors.getDefault();
+  }
+}
+
+```
+
+You can read more about the `@Async` functionality [here](https://www.baeldung.com/spring-async).
+
+### Configuring the Executor
```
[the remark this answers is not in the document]
MatKuhr: I also thought about that, but found that people might not read on and miss the Spring integration, which is probably more relevant to most readers.

docs/java/features/multi-tenancy/thread-context.mdx
```diff
@@ -56,6 +56,98 @@ That also means that the `ThreadContext` will not be used when accessing this in
...
+
+The `ThreadContextExecutors` class leverages a default `ExecutorService` that can be configured.
```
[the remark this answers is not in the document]
MatKuhr: Yes, I only mention `ExecutorService` here to keep it as simple as possible. This is the thing that people are probably most interested in changing (and we do that as well in the example). I figured the detail is not too important and we can just leave the more specific subtype out here.

docs/java/features/multi-tenancy/thread-context.mdx
Johannes-Schneider: Okay, I see your point.

Nevertheless, I think the formulation _"a default executor"_ is not exactly describing what we want to convey here.
At least if I were reading this, I would ask myself _"What is a **default** executor? Is there a non-default one?"_.
Therefore, the actual method names `getDefault` and `setDefault` might not be optimal, as there is nothing else but this single "thingy", right?
At least I would assume if there is a `getDefault` that I somehow can also get a customized instance (similar to what the Apache HttpClient Builder is doing: You may use `HttpClients.default()` or `HttpClients.custom()`).

---

Overall, the most important part here (IMO) is that **all** submitted operations will be executed **in a single instance** of our interface.
And, yes, the instance that should be used to actually run the operations can be changed.
What do you think?

docs/java/features/multi-tenancy/thread-context.mdx
Johannes-Schneider: > We could also just describe both

Yeah, I think we should just mention both usages here.
Besides the technical "downsides" of the suggestion (2), I think it should be mentioned first since the semantic is just crystal clear and it is overall harder to get wrong.
Nevertheless, for more customizability (e.g. changing multiple values at once), suggestion (1) should be preferred.

docs/java/features/multi-tenancy/thread-context.mdx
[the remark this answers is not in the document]
Johannes-Schneider: Okay, fine with me.

docs/java/features/multi-tenancy/thread-context.mdx
```diff
@@ -56,6 +56,106 @@ That also means that the `ThreadContext` will not be used when accessing this in
...
+Runnable myTask = () -> doStuff();
+```
+
+The operation can be executed asynchronously via:
+
+```java
+Future runningTask = ThreadContextExecutors.submit(myTask);
+```
+
+This will use a default `ThreadContextExecutorService` based on a `CachedThreadPool` to create a new Thread and run the task.
```
Johannes-Schneider, line 80: Next attempt for a clearer description 😅

```suggestion
Operations that are submitted this way will be executed by a single instance of `ThreadContextExecutorService`.
By default, this instance is based on a `CachedThreadPool`, which manages the concurrency internally.
```

docs/java/features/multi-tenancy/thread-context.mdx
```diff
@@ -56,6 +56,106 @@ That also means that the `ThreadContext` will not be used when accessing this in
...
+    return ThreadContextExecutors.getDefault();
+  }
+}
+```
+
+You can read more about the `@Async` functionality [here](https://www.baeldung.com/spring-async).
+
+### Configuring the Executor
+
+The `ThreadContextExecutors` class leverages a default `ExecutorService` that can be configured.
```
Johannes-Schneider, line 114:
```suggestion
The `ThreadContextExecutors` class leverages a single `ThreadContextExecutorService` instance that can be configured.
```

docs/java/features/multi-tenancy/thread-context.mdx
```diff
@@ -56,6 +56,106 @@ That also means that the `ThreadContext` will not be used when accessing this in
...
+
+ThreadContextExecutors.submit(myTaskWithTenant);
+```
+
+To avoid multiple wrappings, in particular when changing multiple values, you can also build a custom executor:
+
+```java
+Tenant myTenant = new DefaultTenant("foo");
+ThreadContextExecutor customExecutor = new ThreadContextExecutor()
+            .withListeners(new TenantThreadContextListener(myTenant));
```
Johannes-Schneider, line 154: Should we maybe modify a second value to make the example a little more complete?

docs/java/features/multi-tenancy/thread-context.mdx
```diff
@@ -56,6 +56,114 @@ That also means that the `ThreadContext` will not be used when accessing this in
...
+Below configuration will ensure all methods annotated with `@Async` will have the `ThreadContext` available:
+
+```java
+@EnableAsync
+@Configuration
+public class AsynchronousConfiguration implements AsyncConfigurer {
+
+  @Override
+  public Executor getAsyncExecutor() {
+    return ThreadContextExecutors.getDefault();
```
CharlesDuboisSAP, line 106:
```suggestion
    return ThreadContextExecutors.getExecutor();
```

CharlesDuboisSAP (changes requested):

Johannes-Schneider (commented):
Aside from what @CharlesDuboisSAP suggested: LGTM!

PR #960, opened 2022-08-15
fix: update broken links and throw errors when a broken link is found

tomfrenken:
## What Has Changed?

With this PR, docusaurus will throw an error when a broken link is introduced.
Additionally, some broken links were updated.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -20,7 +20,7 @@ module.exports = {
     'The one-stop shop for developing and extending SAP applications in the cloud.',
   url: 'https://sap.github.io',
   baseUrl: '/cloud-sdk/',
-  onBrokenLinks: 'warn',
+  onBrokenLinks: 'throw',
   onBrokenMarkdownLinks: 'warn',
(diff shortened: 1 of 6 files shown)


docusaurus.config.js
```diff
@@ -20,7 +20,7 @@ module.exports = {
     'The one-stop shop for developing and extending SAP applications in the cloud.',
   url: 'https://sap.github.io',
   baseUrl: '/cloud-sdk/',
-  onBrokenLinks: 'warn',
+  onBrokenLinks: 'throw',
   onBrokenMarkdownLinks: 'warn',
```
deekshas8, line 24: [q] Can we also set `onBrokenMarkdownLinks` to throw so broken markdown links are also covered?

tomfrenken: I guess that would also make sense 🤔

tomfrenken: I've added it ✅

deekshas8 (approved):
LGTM.

PR #964, opened 2022-08-16
docs: Fix images and remove mermaid

marikaner:
This replaces the currently broken mermaid charts with images and fixes other broken images in the JS overview.


florian-richter (approved):
Much better than broken diagrams

PR #966, opened 2022-08-18
Java/v4/mention SJB bom

cschubertcs:
## What Has Changed?

In V4 we introduced a new BOM that combines the SAP Java Buildpack BOM for TomEE, as well as our `sdk-bom`.
This allows consumers to easily consume a tested combination of dependency versions and scopes.

Especially in the archetypes that requires some changes to be able to run the application locally.
This required now some changes on our documentation.

## Manual Checks?


docs-java/getting-started.mdx
```diff
@@ -145,6 +145,40 @@ skipUsageAnalytics: false
 To change the Java version modify the `<java.version>` property in the root `pom.xml`.
 :::
 
+### Run App locally
```
MatKuhr, line 148:
```suggestion
### Run App Locally
```

docs-java/getting-started.mdx
```diff
@@ -145,6 +145,40 @@ skipUsageAnalytics: false
...
+cd application/
+mvn spring-boot:run
+```
+
+</TabItem>
+
+<TabItem value="tomee">
+
+```bash
+mvn clean install -Ddeployment=local
```
MatKuhr, line 170: is the flag actually required when installing?

docs-java/guides/dependencies.mdx
```diff
@@ -196,42 +199,92 @@ Still, the same strategy is also valid and useful for the community Buildpack.
 
 ### Avoiding Scope Conflicts
 
-We recommend overriding the scope for selected dependencies when using the SAP Java Buildpack.
+We recommend using the BOMs provided by the SAP Java Buildpack.
```
MatKuhr, line 202: We can [link here for docs](https://help.sap.com/docs/BTP/65de2977205c403bbc107264b8eccf4b/83d241613dcd41c99ccc308ed0d26399.html).

MatKuhr: And also here: https://help.sap.com/docs/BTP/65de2977205c403bbc107264b8eccf4b/6c6936e8e4ea40c9a9a69f6783b1e978.html

docs-java/guides/dependencies.mdx
```diff
@@ -196,42 +199,92 @@ Still, the same strategy is also valid and useful for the community Buildpack.
...
+This will set the version and the scope of the dependencies the SAP Java Buildpack adds to your application at runtime.
+The scope of the those dependencies is set to `provided`.
 
-- `com.sap.cloud.security:java-api`
-- `com.sap.cloud.security:java-security`
-- `com.sap.cloud.security.xsuaa:token-client`
+Depending on the target runtime defined in your `manifest.yml` you should import one of the following dependencies in your `<dependencyManagement>` section:
 
-You can override the scope by setting including these dependencies explicitly in your `pom.xml` with scope `provided`.
+<Tabs defaultValue="tomcat" values={[
```
MatKuhr, line 209: Not sure if we really want to document to this degree of detail here, since this might change without us noticing. This information is also available on the docs I linked above (they are the same as in oour internal GH)

I think here it's enough to recommend checking the Buildpack BOMs for the exact versions and scopes of dependencies it adds. Also, we could link to the Buildpack group ID on maven central to point people to the latest verions

MatKuhr (commented):
otherwise lgtm except for the branch ofc :D

docs-java/getting-started.mdx
cschubertcs: Yes, as the war created by `install` already contains all `compile` and `runtime` dependencies, but not the `provided` ones.

So, actually, this is the more important flag. The on the `tomee:run` ist just there to make sure that users are aware of the profile in the first place.

PR #968, opened 2022-08-18
Java/actually v4/mention sjb bom

cschubertcs:
## What Has Changed?

In V4 we introduced a new BOM that combines the SAP Java Buildpack BOM for TomEE, as well as our `sdk-bom`.
This allows consumers to easily consume a tested combination of dependency versions and scopes.

Especially in the archetypes that requires some changes to be able to run the application locally.
This required now some changes on our documentation.

## Manual Checks?

### DIFF
--- a/docs-java/guides/dependencies.mdx
+++ b/docs-java/guides/dependencies.mdx
@@ -210,7 +210,7 @@ Depending on the [target runtime](https://help.sap.com/docs/BTP/65de2977205c403b
 
 You can verify the behavior by searching for the scope of, for example, the `com.sap.cloud.security.xsuaa:token-client` dependency in your application.
 
-When you use the BOM of the SAP Java Buildpack, it should have the scope `provided`.
+When you use the BOM of the SAP Java Buildpack the `token-client`  should have the scope `provided`.
 If you are not using the BOM, it should have the scope `compile` or `test`.
 
 :::tip
--- a/docs-java/guides/dependencies.mdx
+++ b/docs-java/guides/dependencies.mdx
@@ -200,6 +200,7 @@ Still, the same strategy is also valid and useful for the community Buildpack.
 ### Avoiding Scope Conflicts
 
 We recommend using the [BOMs provided by the SAP Java Buildpack](https://help.sap.com/docs/BTP/65de2977205c403bbc107264b8eccf4b/6c6936e8e4ea40c9a9a69f6783b1e978.html).
+In our `scp-cf-tomee` Archetype we are using the `sdk-buildpack-bom` which itself depends on the SAP Java Buildpack BOM.
 
 This will set the version and the scope of the dependencies the SAP Java Buildpack adds to your application at runtime.
 The scope of the those dependencies is set to `provided`.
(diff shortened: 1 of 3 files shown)


newtork (approved):

docs-java/guides/dependencies.mdx
```diff
@@ -196,42 +199,33 @@ Still, the same strategy is also valid and useful for the community Buildpack.
 
 ### Avoiding Scope Conflicts
 
-We recommend overriding the scope for selected dependencies when using the SAP Java Buildpack.
+We recommend using the [BOMs provided by the SAP Java Buildpack](https://help.sap.com/docs/BTP/65de2977205c403bbc107264b8eccf4b/6c6936e8e4ea40c9a9a69f6783b1e978.html).
```
MatKuhr, line 202: This is slightly confusing because we later only mentioned our "repackaged" BOM. Maybe we should clarify here that you can use our repackaged one or the SJB BOM directly, if you want to (e.g. because you want to manage the SJB version yourself)

MatKuhr (commented):
otherwise lgtm

docs-java/guides/dependencies.mdx
cschubertcs: The "problem" is from my point of view, that our "special" BOM only relates to `tomee7`, whereas the rest of the part here relates to basically `tomee`, `tomee7`, and `tomcat`.

That's why I tried to mention the special case at the very end.

MatKuhr: hmm that makes sense, but since we ship our BOM as default that should be our recommendation, we should always ship by default what we recommend. So I would just clarify that we already implement the recommendatoiun of using the SJB BOM  through our BOM

cschubertcs: Added a sentence here to make it clear what we are using.

docs-java/guides/dependencies.mdx
```diff
@@ -196,42 +199,34 @@ Still, the same strategy is also valid and useful for the community Buildpack.
...
+In our `scp-cf-tomee` Archetype we are using the `sdk-buildpack-bom` which itself depends on the SAP Java Buildpack BOM.
```
MatKuhr, line 203:
```suggestion
In our `scp-cf-tomee` Archetype we are using the `sdk-buildpack-bom` which itself brings in the SAP Java Buildpack BOM.
```

sounds good, but now that I read it; I think its also worth highlighting the advantage this brings: it ensures SDK and SJB versions are compatible, and you only need to update the SDK version

docs-java/guides/dependencies.mdx
```diff
@@ -196,42 +199,34 @@ Still, the same strategy is also valid and useful for the community Buildpack.
...
-    <scope>provided</scope>
-</dependency>
-<dependency>
-    <groupId>com.sap.cloud.security.xsuaa</groupId>
-    <artifactId>token-client</artifactId>
-    <version>latest-version-here</version>
-    <scope>provided</scope>
-</dependency>
-```
+When you use the BOM of the SAP Java Buildpack, it should have the scope `provided`.
```
MatKuhr, line 212:
```suggestion
When you use the BOM of the SAP Java Buildpack the `token-client`  should have the scope `provided`.
```

Otherwise people may read this as `it` referring to the SJB BOM, which needs scope `import` instead

docs-java/guides/dependencies.mdx
cschubertcs: How about now?

MatKuhr (approved):

PR #971, opened 2022-08-22
chore: adjust proxy settings

FrankEssenberger:
## What Has Changed?

We do not add the proxy settings to the env destination anymore. This is done to the register destination now, the env is taken as is, which is on purpose to not change a destination.

This PR adjust the docu for the BAS where you have to set the proxy now.


docs-js/guides/bas-external-system.mdx
```diff
@@ -169,11 +169,20 @@ If you do not have a `.env` file, create one or adjust the existing one.
...
-destinations="[{"name": "<destinationName>", "url": "http://<destinationName>.dest"}]"
+destinations="[{
+"name": "<destinationName>",
+"url": "http://<destinationName>.dest",
+"proxyConfiguration":{
+      "host": "127.0.0.1",
+      "port": 8887,
+      "protocol":"http"
+  }
+}]"
```
jjtang1985, line 180: [q] `http://<destinationName>.dest` has `<destinationName>` as a placeholder and `.dest` like a pattern.
If my destination name is "my-dest", should I add a url "http://my-dest.dest"?

[req] I guess this is just an example, as we see the hardcoded proxy config.
Please add the example "keyword" explicitly.

jjtang1985 (approved):
Looks good.
I left some comments for your consideration.

docs-js/guides/bas-external-system.mdx
FrankEssenberger: yes it would be "http://my-dest.dest/

PR #989, opened 2022-08-31
fix: add prettier to js doc

FrankEssenberger:
We missed to add the prettier to our `doc-js` folder.


tomfrenken (approved):
LGTM but @newtork take a look if you would also like to be included in the linting, in that case your docs folder would have to be added in the package.json

PR #992, opened 2022-09-01
fix: fix glob pattern

deekshas8:
## What Has Changed?

This PR fixes the prettier glob pattern and adjusts it to run on windows

## Manual Checks?

- [ ] Verify links still work, e.g., if `id` or name of a file is changed.


FrankEssenberger (approved):
LGTM - consider @ZhongpinWang comment on the use `.` in the command and exclude the java and unwanted parts.

PR #994, opened 2022-09-01
Improve http client docs

fwilhe:
Addressed review comments by @marikaner


docs-js/features/connectivity/generic-http-client.mdx
```diff
@@ -114,41 +114,46 @@ The `executeHttpRequest` function returns a promise of <LinkToLatestJsApiDocs sl
 The HTTP client provided by SAP Cloud SDK allows to automatically retrieve a CSRF token for HTTP APIs which support this.
 Fetching CSRF token is automatically enabled for all requests that typically might change data, such as `PUT`, `POST`, `DELETE` or `PATCH` requests, and is not available for `GET` requests.
 
-Fetching the CSRF token can be enabled or disabled via the `options` parameter of `executeHttpRequest`:
+Fetching the CSRF token can be enabled or disabled via the `httpRequestOptions` parameter of `executeHttpRequest`:
```
ZhongpinWang, line 117:
```suggestion
Set `httpRequestOptions` parameter of `executeHttpRequest` to `false` to disable the CSRF token fetching:
```

docs-js/features/connectivity/generic-http-client.mdx
```diff
@@ -114,41 +114,46 @@ The `executeHttpRequest` function returns a promise of <LinkToLatestJsApiDocs sl
...
 
 ```ts
 {
-  fetchCsrfToken: true,
+  fetchCsrfToken: BOOLEAN_TRUE_OR_FALSE;
```
ZhongpinWang, line 121: [pp] I think this code block is unnecessary and can be removed since the lines above and below are clear enough.

docs-js/features/connectivity/generic-http-client.mdx
```diff
@@ -114,41 +114,46 @@ The `executeHttpRequest` function returns a promise of <LinkToLatestJsApiDocs sl
...
 }
 ```
 
+For example, to avoid CSRF token fetching when doing a `PUT`, `POST`, `DELETE` or `PATCH` request, disable it like in this example:
```
ZhongpinWang, line 125: [pp] Also remove this line as it has duplicated meaning if we adopt the suggestion.

docs-js/features/connectivity/generic-http-client.mdx
```diff
@@ -114,41 +114,46 @@ The `executeHttpRequest` function returns a promise of <LinkToLatestJsApiDocs sl
...
+
+```ts
+const response = await executeHttpRequest(
+  {
+    url: `https://httpbin.org/post`
```
ZhongpinWang, line 130: [pp] I would avoid using a 3rd party website here.

docs-js/features/connectivity/generic-http-client.mdx
```diff
@@ -114,41 +114,46 @@ The `executeHttpRequest` function returns a promise of <LinkToLatestJsApiDocs sl
...
+  {
+    method: 'post'
+  },
+  {
+    fetchCsrfToken: false
+  }
+);
+```
+
+In this example, we can only do the request when `fetchCsrfToken` is set to `false` because the httpbin endpoint does not implement CSRF token fetching.
```
ZhongpinWang, line 141:
```suggestion
Disable the CSRF token fetching if the endpoint does not support it.
```
Make it more general (or perhaps remove it since people should probably already know if csrf is supported.)

docs-js/features/connectivity/generic-http-client.mdx
```diff
@@ -114,41 +114,46 @@ The `executeHttpRequest` function returns a promise of <LinkToLatestJsApiDocs sl
...
+For an example with CSRF token fetching enabled see the [http-client.spec.ts](https://github.com/SAP-samples/cloud-sdk-js/blob/main/samples/http-client-examples/http-client.spec.ts) example.
```
ZhongpinWang, line 142: [req] Reformulate the sentence to avoid using the word `examples` too often.

docs-js/features/connectivity/generic-http-client.mdx
```diff
@@ -158,30 +163,28 @@ const myCustomParameterEncodingFunction: ParameterEncoder = function (
...
   }
 
   return encodedParams;
 };
 
-const encodingResponse = await executeHttpRequest(
+const response = await executeHttpRequest(
   {
-    url: 'http://localhost:3000/encoding'
+    url: 'https://httpbin.org/anything'
```
ZhongpinWang, line 174: Same as above. Avoid using 3rd party URL. Use `example.com` instead as it is [reserved from IANA](https://en.wikipedia.org/wiki/Example.com).

docs-js/features/connectivity/generic-http-client.mdx
```diff
@@ -211,13 +214,12 @@ Headers and query parameters with the same name are prioritized in the following
 In this example, the `custom` `apiKey` and `myParam` override the request config:
 
 ```ts
-executeHttpRequestWithOrigin(
+const response = await executeHttpRequestWithOrigin(
   {
-    url: 'http://localhost:3000/'
+    url: 'https://httpbin.org/anything'
```
ZhongpinWang, line 219: Same as above.

ZhongpinWang (commented):
The example part for CSRF token could be simplified as examples for true and false are essentially the same.
Also, I personally prefer not to use any 3rd URL in the example but rather use `example.com`.

fwilhe:
> Also, I personally prefer not to use any 3rd URL in the example but rather use example.com.

I see your point, but my reasoning is that this allows us to have actually working examples which users can copy/paste, and wich we can have working in tests in the samples repo. Unless there are any legal issues with that (which I'm not aware of) I'd like to keep it.

docs-js/features/connectivity/generic-http-client.mdx
fwilhe: No this is not correct because the `fetchCsrfToken` field of the `httpRequestOptions` parameter needs to be set to `false`.

docs-js/features/connectivity/generic-http-client.mdx
fwilhe: I think this helps with clarity and I'd like to leave it in.

docs-js/features/connectivity/generic-http-client.mdx
fwilhe: I think this helps with clarity and I'd like to leave it in.

docs-js/features/connectivity/generic-http-client.mdx
fwilhe: I see your point, but my reasoning is that this allows us to have actually working examples which users can copy/paste, and wich we can have working in tests in the samples repo. Unless there are any legal issues with that (which I'm not aware of) I'd like to keep it.

docs-js/features/connectivity/generic-http-client.mdx
fwilhe: I think this sentence explains what's happening in the example and it adds value imo, so I'd like to leave it.

fwilhe:
Addressed all comments by @ZhongpinWang

marikaner (dismissed):

marikaner (approved):
LGTM

PR #1004, opened 2022-09-09
Remove pre-release info on sdk v2

fwilhe:
Those pages are not accurate anymore so they might lead to confusion.


florian-richter (approved):
[q] Should the upgrade guide be part of the releases folder?

fwilhe:
> [q] Should the upgrade guide be part of the releases folder?

Merged, can be addressed in another PR if necessary.

PR #1006, opened 2022-09-09
Document EDM to TypeScript type conversion and limitations

fwilhe:
## What Has Changed?

Documents EDM data types and how they are converted by the default (de)serializers with known limitations.

## Manual Checks?

- [ ] ~~Update the [feature matrix](https://github.com/SAP/cloud-sdk/blob/main/docs/components/data/features.js), e.g., if a new feature is added.~~


docs-js/features/odata/common/edm.mdx
```diff
@@ -0,0 +1,75 @@
...
+---
+
+## Type conversions
+
+The SAP Cloud SDK for JavaScript automatically converts values encoded in EDM to respective types in JavaScript and TypeScript.
+Some types are specific to an OData version.
+
+For definition of the EDM data types in OData, see [the v2 spec](https://www.odata.org/documentation/odata-version-2-0/overview/) or the [v4 spec](http://docs.oasis-open.org/odata/odata/v4.0/os/part3-csdl/odata-v4.0-os-part3-csdl.html#_The_edm:Documentation_Element) of OData.
+
+The SAP Cloud SDK converts EDM types into TypeScript types as described by the following tables.
```
florian-richter, line 24: It should be mentioned that this applies to the default deserializers and can be modified.

docs-js/features/odata/common/edm.mdx
```diff
@@ -0,0 +1,75 @@
...
+| `Edm.DateTimeOffset` | `Moment`        |
+| `Edm.Binary`         | `string`        |
+| `Edm.Any`            | `any`           |
+
+### OData v2 specific
+
+| EDM Type       | TypeScript Type |
+| -------------- | --------------- |
+| `Edm.DateTime` | `Moment`        |
+| `Edm.Time`     | `Time`          |
```
florian-richter, line 51: I think it would be helpful to mention that Time is a custom type from us which is this:

```
/**
 * Interface to represent Time or Duration.
 */
export interface Time {
  /**
   * The number of hours.
   */
  hours: number;
  /**
   * The number of minutes.
   */
  minutes: number;
  /**
   * The number of seconds.
   */
  seconds: number;
}
```

Either a short description or a link to our API reference for this interface would be nice.

docs-js/features/odata/common/edm.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+:::caution Precision of Conversion
+
+While converting `Edm.DateTime` to `Moment` the smallest considered unit is milliseconds.
```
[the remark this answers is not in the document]
fwilhe: Looks like it supports 7 decimal places for fractions of a second (for whatever erp use-case this might be useful..)

Screenshot 2022-09-12 at 09 20 08

docs-js/features/odata/common/edm.mdx
```diff
@@ -0,0 +1,78 @@
...
+  - TypeScript
+  - client
+  - generate
+---
+
+import LinkToLatestJsApiDocs from '@site/src/sap/sdk-js/LinkToLatestJsApiDocs';
+
+## Type Conversions
+
+The SAP Cloud SDK for JavaScript automatically converts values encoded in EDM to respective types in JavaScript and TypeScript.
```
marikaner, line 21: [q] Are those values really encoded?
[q] When does the conversion happen? I think it "only" happens when deserializing entities / complex types / primitive types after calling `execute`.

docs-js/features/odata/common/edm.mdx
```diff
@@ -0,0 +1,78 @@
...
+| `Edm.Single`         | `number`        |
+| `Edm.Double`         | `number`        |
+| `Edm.Float`          | `number`        |
+| `Edm.Byte`           | `number`        |
+| `Edm.SByte`          | `number`        |
+| `Edm.DateTimeOffset` | `moment.Moment` |
+| `Edm.Binary`         | `string`        |
+| `Edm.Any`            | `any`           |
+
+### OData v2 Specific
```
marikaner, line 49: [pp] If you add "Types" I find it more consistent.
```suggestion
### OData v2 Specific Types
```

docs-js/features/odata/common/edm.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+| EDM Type        | TypeScript Type                                                                         |
+| --------------- | --------------------------------------------------------------------------------------- |
+| `Edm.Date`      | `moment.Moment`                                                                         |
+| `Edm.Duration`  | `moment.Duration`                                                                       |
+| `Edm.TimeOfDay` | <LinkToLatestJsApiDocs slug="interfaces/sap_cloud_sdk_odata_common.Time" name="Time" /> |
+
+:::caution Precision of Conversion
+
+While converting `Edm.DateTime` to `moment.Moment` the smallest considered unit is milliseconds.
```
marikaner, line 74: [req] Copy paste error
[q/pp] This error happened because it is duplicated text. I wonder whether we could find a way to put this together with the OData v2 warning. Did you consider this somehow? 
```suggestion
While converting `Edm.DateTime` to `moment.Moment` the smallest considered unit is milliseconds.
```

docs-js/features/odata/overview.mdx
```diff
@@ -30,6 +30,13 @@ There are two widely used versions of the protocol: OData v2 and OData v4.
 The latter comes with more and improved functionality.
 Full details on the standards are to be found in the [documentation](https://www.odata.org/documentation/).
 
+### Entity Data Model (EDM)
+
+OData makes use of the Entity Data Model (EDM).
+The SAP Cloud SDK for JavaScript automatically converts values encoded in EDM to respective types in JavaScript and TypeScript.
+
+See the [page on edm](./common/edm.mdx) for details and limitations of that conversion.
```
marikaner, line 38: [pp] I am torn whether it is a good idea to have more pages also for minor topics or not and where to draw the line. For comparison, I looked at the docusauraus docs and I think we are too broad when it comes to the general odata client description, but just edm seems to fine granular.
I think I would prefer something like Serialization and Deserialization as a topic, where EDM is part of it.
However, doing that would be out of scope for this issue...

docs-js/features/odata/common/edm.mdx
marikaner: [nth/q] Consider the following case: A user just looks at the OData v4 section of this article. This suggests, that we only support 3 EDM types. Maybe we should link to the common parts as well?

marikaner (changes requested):
I like the structure and have mostly minor feedback + a question whether to put this into its own page.

docs-js/features/odata/common/edm.mdx
fwilhe: Not sure if we have the same understanding of "encoding" but I would say that every computer representation of values is a type of "encoding", also ascii or unicode.

EDM defines how a date or a number is represented and that representation is "encoding" in my understanding. Also seems to align with [wikipedia's definition](https://en.wikipedia.org/wiki/Code).

But if you have an alternative term to use please suggest that.

marikaner: It really was meant as a question, because I imagine url encoding or base 64 encoding when I read that. The word I would suggest would be something in the direction of "represented", but you can just decide what you like most.

marikaner (approved):
👍 Love it!

docs-js/features/odata/execute-odata-request.mdx
```diff
@@ -145,6 +146,43 @@ The last step when making a request using the SAP Cloud SDK is the request execu
...
 
+### Conversion of OData Types
+
+OData uses the Entity Data Model (EDM) to represent values such as strings, numbers, dates and times.
+Calling the `execute` function by default automatically converts OData Entity Data Model (EDM) values to their respective types in JavaScript and TypeScript.
+
+By default, the SAP Cloud SDK converts EDM types into TypeScript types as described by the following tables.
+This can be changed by providing custom serializers and deserializers as described [here for OData v2](./odata-v2-client.mdx#customize-de-serialization) and [here for OData v4](./odata-v4-client.mdx#customize-de-serialization).
+
+For definition of the EDM data types in OData, see [the v2 spec](https://www.odata.org/documentation/odata-version-2-0/overview/) or the [v4 spec](http://docs.oasis-open.org/odata/odata/v4.0/os/part3-csdl/odata-v4.0-os-part3-csdl.html#_The_edm:Documentation_Element) of OData.
```
florian-richter, line 157:
```suggestion
For definition of the EDM data types in OData, see [the v2 spec](https://www.odata.org/documentation/odata-version-2-0/json-format/) or the [v4 spec](http://docs.oasis-open.org/odata/odata/v4.0/os/part3-csdl/odata-v4.0-os-part3-csdl.html#_The_edm:Documentation_Element) of OData.
```
[pp] You may want to link the OData-JSON spec or the OData spec overview page instead

florian-richter (commented):
Some spelling and comma findings. Otherwise looks good!

docs-js/features/odata/execute-odata-request.mdx
```diff
@@ -145,6 +146,43 @@ The last step when making a request using the SAP Cloud SDK is the request execu
 Once the request is configured use the `execute` function and pass a [destination](../connectivity/destination) to it.
 This will automatically construct your request URL and headers, execute the request, and return a typed response if applicable.
 
+### Conversion of OData Types
+
+OData uses the Entity Data Model (EDM) to represent values such as strings, numbers, dates and times.
```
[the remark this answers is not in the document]
fwilhe: 🇬🇧

PR #1030, opened 2022-09-27
Docs style guide

deekshas8:
## What Has Changed?

Added the formatting style guide.

## Manual Checks?


style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+
+Use code format for text that should be entered by user
+
+✅ In the **Name** field, enter `my-destination`
+
+❌ In the **Name** field, enter my-destination
+
+## Avoid Using We
+
+✅ More than 100 deprecated methods have been removed.
```
FrankEssenberger, line 141: [q] how do you avoid `we` and also use active language?

style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+
+❌ placeholderName
+
+<!-- vale on -->
+
+### User-Replaced Placeholders in Inline Text
+
+When placeholders occur in a sentence use code format in italics.
+
+✅ where _`YOUR_APPLICATION_NAME`_ is replaced by the name you specified.
```
FrankEssenberger, line 36: general: I thought we should use active language. If so would be nice if the examples feature that.

FrankEssenberger (approved):
LGTM only minor question regarding active/passive vioce.

style-guide.md
deekshas8: @FrankEssenberger I used the examples from our text for the formatting so it was not super generic examples but provided more clarity. Since this guide was supposed to be strictly for formatting (scope of the ticket) and not about language and writing style, I decided not to go there.

I can however add a small section still with some examples.

style-guide.md
deekshas8: In some cases, there can be exceptions as explained [here](https://developers.google.com/style/voice#exceptions) (To de-emphasize a subject or actor.).

FrankEssenberger: ok, this is good to know. For me as a non native speaker it is hard to judge when it is ok to use passiv form, but this is the problem with normal language with their non strict rules

style-guide.md
```diff
@@ -0,0 +1,172 @@
+# About This Guide
+
+This style guide covers some general text-formatting conventions that should be applied when writing documentation for the SAP Cloud SDK.
+
+## Formatting
```
marikaner, line 5: [req] The highest level title should be the title of the document. Currently "Formatting" us a section under "About This Guide".
Suggestion: remove the first level and drag all sections one level up.
```suggestion
# Formatting

This style guide covers some general text-formatting conventions that should be applied when writing documentation for the SAP Cloud SDK.
```

style-guide.md
```diff
@@ -0,0 +1,172 @@
+# About This Guide
```
marikaner, line 1: [pp] I think it would be nice to be consistent with the naming of our files. In the JS repo, we call it `STYLEGUIDE.md`.

style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+
+❌ Placeholder-Name
+
+<!-- vale off -->
+
+❌ placeholderName
+
+<!-- vale on -->
+
+### User-Replaced Placeholders in Inline Text
```
marikaner, line 32: [req] I find it a bit confusing that this section and the next section are on the same level as the general place holder section. Suggestion:
Have 1 section on "Placeholders" with some general guidance and below that add 3 sections on:
- Placeholders in Code
- Placeholders in Text
- Placeholder Text (or Placeholder Syntax, but I slightly lean towards Text)

To me that would be the natural order of how I think about that. I first have a code example. Then I have the text that I use to explain the example. From the examples in the guide I probably already have an idea of how the placeholder text should look like, but if I want more details, there is a detailed section on this.

style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+
+❌ PLACEHOLDER-Name
+
+❌ Placeholder-Name
+
+<!-- vale off -->
+
+❌ placeholderName
+
+<!-- vale on -->
```
marikaner, line 30: I think we should also include the rule that possessive pronouns should not be included.

style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+
+❌
+
+```YAML
+applications:
+  - name: <YOUR-APPLICATION-NAME>
+    path: .
+```
+
+### Software Product and their Command Line Utility
```
marikaner, line 66: [nth]
```suggestion
### Software Products and Command Line Utilities
```

style-guide.md
marikaner: [nth] I think it would be helpful to have a table of contents same as here: (https://github.com/SAP/cloud-sdk-js/blob/main/STYLEGUIDE.md#table-of-contents)

style-guide.md
```diff
@@ -0,0 +1,172 @@
+# About This Guide
+
+This style guide covers some general text-formatting conventions that should be applied when writing documentation for the SAP Cloud SDK.
+
+## Formatting
+
+### Code Format for Inline Code
+
+Use code font to mark up code, such as class names, method names, HTTP status codes/verbs/content-types, env variables, console output, etc.
```
marikaner, line 9: [req] I think this should be a list:
```suggestion
Use code font to mark up code, such as:
- class names
- method names
- HTTP status codes/verbs/content-types
- env variables
- console output
- etc.
```

style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+
+✅ instance of the `BusinessPartner` entity class
+
+For a complete list, see [here](https://developers.google.com/style/code-in-text#some-specific-items-to-put-in-code-font)
```
marikaner, line 13: [req] I think this section would benefit from a subsection what not to put into code font.

style-guide.md
marikaner: [pp] I am not sure that these two things need to be put together. Command line utilities could just be mentioned in the "Code Format for Inline Code" or be a separate section.

style-guide.md
marikaner: [pp] But I agree, we could easily change this example to active voice without making this part of the guide.
```suggestion
✅ Replace _`YOUR_APPLICATION_NAME`_ with your application name.
```

marikaner: [req] However, according to the google guide (and I like this rule) we should not use possessive pronouns, so maybe this example should not use a possessive pronoun either ;)

style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+Use code font.
+
+✅ against a destination with URL `https://my.s4-system.com`
+
+### Filenames and FilePaths in Text
+
+Use code font.
+Include the word file after the filename.
+
+✅ Add your own `package.json` file.
```
marikaner, line 85: [nth] Maybe add an example for a file path as well?

style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+
+### File Type in Text
+
+✅ metadata specification in the EDMX format
+
+❌ metadata specification in the `EDMX` format
+
+Try to use the formal file type name, not the filename extension.
```
marikaner, line 93: [req] Can you give an example for this? e.g. XML file vs. .xml file?

style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Error Messages in Text
+
+Sentence-style capitalization.
+Enclose error messages in quotation marks when referring to them in text.
```
marikaner, line 98: [q] I think there can be quite complex cases where the error already includes both " and '. How should this be handled?

style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+
+✅ If you see the error "Module not found: Error: Can't resolve 'https' in"
+
+❌ If you see the error `Module not found: Error: Can't resolve 'https' in`
+
+### Example Domains
+
+For generic domain names in examples, use `example.com`, `example.org`, or `example.net`.
```
marikaner, line 106: [q] Should we restrict this to only `example.com`? I doubt that all three will be needed.
[q] Also, what do I do if I need to use two different domains. I think `example.com` are `example.org` too similar. I guess then something like `other.domain.com` would be OK (or would it not?). Should we document this or rather not?
[q] Strictly speaking this is not about formatting, but naming. Should we move this down to the other non-formatting related aspects?

style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Graphical User Interface Elements
+
+Bold all graphical user interface element names, including, menus, button, dialog box, list items, and window.
```
marikaner, line 110: [req] I think this should be a list for better readability.

style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+
+✅ On the **Settings** page, click **Show API Key**
+
+❌ On the settings page, click _Show API Key_.
+
+✅ In the dropdown, choose **BasicAuthentication** as the authentication type
+
+❌ In the dropdown, choose `BasicAuthentication` as the authentication type
+
+Don't include the type of UI element, such as button or checkbox, unless it adds needed clarity.
```
marikaner, line 121: [pp] I think we should either drag this text up to the top of this section or introduce subsections for better overview. I don't know what the best option here is...

style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+Use sentence-style capitalization unless you need to match the UI.
+If a label ends with a colon/an ellipsis, don't include that end punctuation.
+
+✅ To start your application, click **Launch**.
+
+❌ You can start your application via the "launch.." button.
+
+❌ You can start your application via the "LAUNCH" button.
```
marikaner, line 129: [req] I think these examples are not enough to understand the requirements well. The main problem is, that we do not see any UI with a button, therefore adding info on the button label would be helpful.
```suggestion
Button label: launch...

✅ To start your application, click **launch**.

✅ **Launch** starts your application.

❌ To start your application, click **launch...**.

❌ To start your application, click **LAUNCH**.

❌ You can start your application via the "launch.." button.
```

style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+✅ On the **Settings** page, click **Show API Key**
+
+❌ On the settings page, click _Show API Key_.
+
+✅ In the dropdown, choose **BasicAuthentication** as the authentication type
+
+❌ In the dropdown, choose `BasicAuthentication` as the authentication type
+
+Don't include the type of UI element, such as button or checkbox, unless it adds needed clarity.
+Use sentence-style capitalization unless you need to match the UI.
```
marikaner, line 122: [q] When do I not need to match the UI? I really don't understand when this would be the case.

style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+
+❌ This mapping applies to the move to open source version 1.X.Y..
+
+<!-- vale on -->
+
+Avoid words like "currently" and "new".
+
+✅ In version 2, we allow to specify the origin of the header
+
+❌ In the new version we allow to specify the origin of the header
```
marikaner, line 165: [req] Remove "we" from these samples ;)

style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+
+### User Entered Text
+
+Use code format for text that should be entered by user
+
+✅ In the **Name** field, enter `my-destination`
+
+❌ In the **Name** field, enter my-destination
+
+## Avoid Using We
```
marikaner, line 139: [q] I think from here on this is not formatting anymore, maybe this should have another separate heading like "Style"?

style-guide.md
```diff
@@ -0,0 +1,172 @@
...
+## Formatting
+
+### Code Format for Inline Code
+
+Use code font to mark up code, such as class names, method names, HTTP status codes/verbs/content-types, env variables, console output, etc.
+
+✅ instance of the `BusinessPartner` entity class
+
+For a complete list, see [here](https://developers.google.com/style/code-in-text#some-specific-items-to-put-in-code-font)
+
```
marikaner, line 14: [req] I think we should have a subsection here for the special case of methods (and functions) => https://developers.google.com/style/code-in-text#method-names
[nth] Maybe it would also be nice to have non-formatting related section below on when to speak of methods and when to call something a function. I have the impression that this is still misunderstood a lot.

marikaner (changes requested):
I find this very helpful already. I left many comments on structure and some missing content in my opinion. I think it could be helpful to add a short sentence to every example why it is correct/incorrect, similar to how we did it in the [code style guide](https://github.com/SAP/cloud-sdk-js/blob/main/STYLEGUIDE.md).

style-guide.md
deekshas8: I assume the same. Also I didn't notice any issues in markdown syntax if you add text that includes. When the error message is too complex/large, we can use the  fenced block syntax (which I have seen being used)

style-guide.md
deekshas8: When either the labels themselves are inconsistent or all caps, in that case, change all to sentence style. I improved the text a bit to avoid confusion.

style-guide.md
deekshas8: [1] 👍🏽 
[2] I honestly dont have a good suggestion here. We can leave this point open and document it when we do have such a scenario and have made a decision what to use.  
[3] I agree, but I would keep it here because It also doesn't fit under Language and Style and Punctuation either. Later if we have a general section, we can move it there.

STYLEGUIDE.md
```diff
@@ -0,0 +1,344 @@
...
+- put an empty pair of parentheses after a method name to indicate that it's a method
+
+✅ This can be done by using the `asChildOf()` method
+
+❌ This can be done by using the `.asChildOf` method
+
+#### Function and Method
+
+If a function is part of an instance of a class then it is a method, else it is a function.
+Use those terms accordingly.
```
marikaner, line 67: Thank you 😊

STYLEGUIDE.md
```diff
@@ -0,0 +1,344 @@
...
+
+✅ Functions such as `execute()` and `executeHttpRequest()` call `getDestination()` internally
+
+❌ Methods such as `execute()` and `executeHttpRequest()` call `getDestination()` internally
+
+## Placeholders
```
marikaner, line 73: I find this much more easy to follow now.

STYLEGUIDE.md
```diff
@@ -0,0 +1,344 @@
...
+
+## Filenames and FilePaths
+
+Use code font.
+Use hyphens, not underscores, to separate words.
+Include the word file after the filename.
+
+✅ Add your own `package.json` file.
+
+✅ This creates a basic class inside `src/business-partners` folder.
```
marikaner, line 154:
```suggestion
✅ This creates a basic class inside the `src/business-partners` folder.
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,344 @@
...
+✅ On the **Settings** page, click **Show API Key**
+
+❌ On the settings page, click _Show API Key_.
+
+✅ In the drop-down, select **BasicAuthentication** as the authentication type
+
+❌ In the dropdown, choose `BasicAuthentication` as the authentication type
+
+- When a label is all uppercase, use sentence-style capitalization.
+- If a label ends with a colon/an ellipsis, don't include that end punctuation.
```
marikaner, line 197: [q] I still find this a weird rule. From the google docs they say:

> In most cases, follow the capitalization as it appears on the page. However, if labels are inconsistent or they're all uppercase, use sentence case.

I think the first sentence is important, that most of the time it should be used as it appears on the page. Can we add this?
I personally, don't understand why you would change the capitalization in the documentation even if it is inconsistent or all uppercase. It is fine for me to leave this rule however - I just don't see the point of doing that, because if the button is labeled "REFRESH" the docs can also tell me to click "REFRESH", not "Refresh" - that even seems wrong to me... 

Anyways, how I understand the original docs from google, you should only do this if all labels are uppercase...

¯\\_(ツ)_/¯

STYLEGUIDE.md
```diff
@@ -0,0 +1,344 @@
...
+
+✅ The SAP Cloud SDK does not guarantee any API compatibility for future updates
+
+❌ We do not guarantee any API compatibility for future updates
+
+### Second Person
+
+Use you instead of we.
+When telling the reader to do something, use the imperative (implicit you).
+Avoid your.
```
marikaner, line 275: [req] Could you please add an example for this?

STYLEGUIDE.md
```diff
@@ -0,0 +1,344 @@
...
+## Politeness
+
+Be polite, but using please in a set of instructions is overdoing the politeness.
+
+- Don't use please when explaining a procedure, even when explaining a difficult task.
+- Don't use the phrase **Please note**
+
+❌ Please note that the `CUSTOMER_NAME` and `CUSTOMER_ACCOUNT_GROUP` are properties of the entity
+
+Use please only when asking for permission or forgiveness—for example, when what you're asking for benefits us, inconveniences a reader, or suggests a potential issue with a product.
```
marikaner, line 294: [nth] I think there is a typo in this sentence.
Also, please add spaces around the "-".

marikaner (approved):
I still left some comments, but this already looks very good and sensible to me.

STYLEGUIDE.md
deekshas8: The first example is for 'your' (and also 'we').

✅ Create an Ingress to make the application available from outside the cluster.

❌ To make your application available from outside the cluster we will create an Ingress.

deekshas8: The second example is for the imperative case. Do you think splitting them into subsections (1. Avoid you/your 2. Imperative case) would be better in this case?

STYLEGUIDE.md
deekshas8, line 294:
```suggestion
Use 'please' only when asking for permission or forgiveness. For example, when what you're asking for:
- benefits us
- inconveniences a reader
- or suggests a potential issue with a product
```

STYLEGUIDE.md
deekshas8: I agree. I am also fine if we keep it simple and just say 'follow the capitalization as it appears on the page'. I think it's easier to follow when writing and also for the user.

STYLEGUIDE.md
marikaner: I think it would be best to have one correct and incorrect example for each rule. With that you could make the examples shorter (and easier to grasp). You could even consider adding another layer of subsections for each of those rules (so that it comes up in the TOC).

The reader of this document will most likely not read the document as they would read a blog post. Instead they will rather use this as a reference when writing text.
I really like that you wrote very short text, that makes it easy to find. Still, I think this could potentially be even easier. Currently I still have to read text to get the information and if I'm in the wrong section, I have to read more text.
Side note: I think incorrect examples should come first, because in my brain this triggers an aha-effect.

That said, I think this guide is already very helpful, this might be nitpicky again ;)

Direction, of what I meant:

```ts
### Second Person

#### Don't use "we"

Don't use "we" when addressing the reader.

❌ Next, we will create an Ingress.

✅ Create an Ingress.

#### Use the imperative

Use the imperative (implicit you), when you give instructions to the reader.

❌ Next, you will create an Ingress.

✅ Create an Ingress.

#### Avoid "your"

❌ Start your application.

✅ Start the application.

```

STYLEGUIDE.md
```diff
@@ -0,0 +1,557 @@
...
+    path: .
+```
+
+❌
+
+```YAML
+applications:
+  - name: <YOUR-APPLICATION-NAME>
+    path: .
+```
```
marikaner, line 157: This should probably be switched 😊

PR #1034, opened 2022-09-29
[js] Improve error handling

jjtang1985:
From the user feedback, they did not see the real error message from an `AxiosError`, if they follow our current doc.
I added a section for improvement.


docs-js/features/error-handling.mdx
```diff
@@ -48,3 +48,46 @@ Root cause: Request failed with status code 404
...
+```
+Message: Request failed with status code 503
+Response body: {
+  code: 'SERVICE_UNAVAILABLE',
+  message: 'From XXX until XXX there is a planned downtime for the API sandbox system. During this period you cannot use the Try it Out feature on our API sandbox system to test an API service.'
+}
+```
+
+:::tip
+To display the complete root cause error object, here is an example:
```
FrankEssenberger, line 87: [pp] the `here is an example` sounds not so nice to me , perhaps use `If you want to see the rootCause of an error object add the following log statement:`.

FrankEssenberger (approved):

PR #1036, opened 2022-09-29
[Java V4] Rewrite Service Binding Feature Documentation

Johannes-Schneider:
## What Has Changed?

This PR rewrites the feature documentation about service binding loading to adhere to the new use of the [BTP Environment Variable Access](https://github.com/SAP/btp-environment-variable-access) library.

## Manual Checks?


MatKuhr (commented):
for me the doc is somewhat lacking a use case. In what situation does a user actually need this? Maybe we can just make this internal documentation as well if we don't see enough value

Johannes-Schneider:
@MatKuhr wrote:
> for me the doc is somewhat lacking a use case. In what situation does a user actually need this? Maybe we can just make this internal documentation as well if we don't see enough value

Personally, I'd be fine with not merging this and rather convert this to some sort of internal documentation (if it's needed at all?).

Therefore, I'll close this PR for now.
We can still add documentation about the integration in case customers are wondering.

PR #1037, opened 2022-09-30
[js] Trouble shooting guide

FrankEssenberger:
## What Has Changed?

A trouble shooting guide for connectivity issues is create in this PR. I added the picutre only in light mode. Once all suggestions are in, I will create a dark version to avoid double work. 

Closes https://github.com/SAP/cloud-sdk-backlog/issues/853


docs-js/trouble-shooting.mdx
```diff
@@ -0,0 +1,153 @@
...
+- The user is locked in the target system
+- The permission of the user are insufficient.
+- The scopes of the JWT in the `authTokens` are insufficient.
+
+If your request reaches the target system and gets rejected there the problem lies most likely not with the SAP Cloud SDK.
+The error logs of the target system may provide additional information on the error.
+
+### Activate Backend Logs
+
+You have a problem connecting to an On-premise system via a cloud-connector and connectivity service.
```
deekshas8, line 109:
```suggestion
You have a problem connecting to an On-premise system via a Cloud Connector and connectivity service.
```

Its written as 'Cloud Connector' in the other docs

docs-js/trouble-shooting.mdx
```diff
@@ -0,0 +1,153 @@
...
+Often there is an issue in the cloud-connector setup.
+To see if your request reaches the backend system activate the HTTP logs:
+
+- Go to transaction **SMICM**.
```
deekshas8, line 113: [info] Since the steps sequence matters, Numbered list would be a better option.

docs-js/trouble-shooting.mdx
```diff
@@ -0,0 +1,153 @@
...
+### Use Minimal Example
+
+Often the issue is related to a framework which uses the SAP Cloud SDK and not to the code itself.
+Therefore, a minimal example should be added to the code to ensure the error is related to the SAP Cloud SDK.
+
+```ts
+import { executeHttpRequest } from '@sap-cloud-sdk/http-client';
+import { getDestination } from '@sap-cloud-sdk/connectivity';
+
+const destination = await getDestination({
```
deekshas8, line 47: [nitpick] If I copy this sample as is (with my current proj config) I would get the issue with top level await. Maybe we can have it in a function so no additional effort is needed by the user.

deekshas8 (commented):
Honestly, I found the diagram a bit hard to follow and would be fine even if we don't have it. The content is good. But regarding its structure, I feel it would be more helpful if it's sectioned according to error code/messages that a user would commonly encounter. For example, 'Failed to add proxy authorization header - client credentials grant failed!' and then how to troubleshoot this specific error.

docs-js/trouble-shooting.mdx
FrankEssenberger: so you are not on ESM6 :-). I guess commonJs is still the default so I add it.

docs-js/trouble-shooting.mdx
```diff
@@ -0,0 +1,160 @@
...
+
+### Activate Backend Logs
+
+You have a problem connecting to an On-premise system via a Cloud Connector and connectivity service.
+Often there is an issue in the Cloud Connector setup.
+To see if a request reaches the backend system, activate the HTTP logs.
+
+1. Go to transaction **SMICM**.
+1. Navigate to **GOTO->Trace Level** and set the trace level to 3 and activate HTTP logs.
+   1.You may want to reset the log file to have no old logs **GOTO->Trace File**.
```
fwilhe, line 121:
```suggestion
   1. You may want to reset the log file to have no old logs **GOTO->Trace File**.
```

fwilhe (approved):

marikaner (commented):
There is are typos in the chart: "fix destinataion" => "fix destination"  and "not a SDK issue" => "not an SDK issue".

I also think that this chart is difficult to follow, e.g.:
- there is no activity between "example works?" and "code works?"
- Why is the next step after fixing destination issues to ask for help?
- I think if the main logic would flow from left to right this would be an improvement.

PR #1048, opened 2022-10-07
Add version dropdown for Java v3 and v4

CharlesDuboisSAP:
## What Has Changed?

Both the versioned v3 and current v4 java docs can be displayed using the dropdown.

## Manual Checks?


KavithaSiva (approved):
LGTM. Just not sure if we have to merge this now or only after our release is on Maven central.

@newtork What do you think?

PR #1063, opened 2022-10-13
Document Sentence per Line

fwilhe:
Document that and why we should follow the "one sentence per line" paradigm.


STYLEGUIDE.md
```diff
@@ -317,9 +317,24 @@ Unless a text relates to a specific version, do not use formats x.y or x.y.z
...
 
-✅ In version 2, you can specify the origin of the header
+✅ In version 2, you can specify the origin of the header.
+
+## Paragraphs
+
+In markdown, an empty line denotes the start of a new paragraph.
+Use paragraphs to structure your text into individual units.
+
+### Sentence per Line
```
deekshas8, line 329:
```suggestion
### Use One Sentence per Line
```

STYLEGUIDE.md
```diff
@@ -317,9 +317,24 @@ Unless a text relates to a specific version, do not use formats x.y or x.y.z
...
+
+A line break inside a paragraph is not visible in the rendered text.
+Using newlines after each sentence make it easier to restructure text and to view diffs of prose.
+See also [AsciiDoc Recommended Practices](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line) for a more elaborate reasoning of this practice.
+
+❌ Don't put multiple sentences of prose into one line.
```
deekshas8, line 335: [req] The current examples reiterate the rule, which is clear from the intro text. Adding an incorrect example (best if from the actual docs) where text should have been separated into para's instead of one long para would be more helpful.

deekshas8 (commented):
Please update the TOC as well.

deekshas8 (approved):

PR #1068, opened 2022-10-17
fix: Remove notion of pregenerated clients from summary

marikaner:
I think this was forgotten in the previous update.


FrankEssenberger (approved):
Yes good catch

PR #1071, opened 2022-10-17
[js] fix broken redirection to the latest API doc

shojisoma:
## What Has Changed?

Closes https://github.com/SAP/cloud-sdk-backlog/issues/889
this fixes a broken redirection from a path `/api/latest` to the latest API doc.

## Manual Checks?


src/pages/api/latest.mdx
```diff
@@ -9,8 +9,6 @@ keywords:
   - sap cloud sdk
 ---
 
-import { Redirect } from '@docusaurus/router';
-import useBaseUrl from '@docusaurus/useBaseUrl';
 import jsSdkVersions from '../../../static/api/versions';
-
-<Redirect to={useBaseUrl(`api/${jsSdkVersions[0]}/`)} />
+// prettier-ignore
+window.location.href = `/cloud-sdk/api/${jsSdkVersions[0]}/`
```
marikaner, line 14: @shojisoma Can you explain what this changes?
I think there might be a slight misunderstanding, because this will still go against something like `api/2.9.0` instead of `api/latest`.

shojisoma: Thank you for your review.

To avoid misunderstanding, let me confirm what the ticket should fix. I thought what that has to do is make a redirection to the latest API doc page when a user access `api/latest`. Is that correct?

I found [an docusaurus issue ](https://github.com/facebook/docusaurus/issues/3421) created by you. The problem  is that `Redirect` doesn't work because the API docs are in `static` which are not managed by at all by docusaurus. The `Redirect` cannot render the external pages. Since I saw the issue, I thought what this has to do is fix the `Redirect`.

My update has [a problem](https://github.com/SAP/cloud-sdk/actions/runs/3267466582/jobs/5372719802) while the building process because docusaurus could not render the static page with path `/api/latest.` It works fine on local but got the error after I made PR hence I turned this PR to draft.

PR #1076, opened 2022-10-18
[Java] Update Getting Started Docs

Johannes-Schneider:
## What Has Changed?

This PR updates the Java _Getting Started_ documentation, which still contained a few outdated references.

~Furthermore, it also updates the instructions on how to run an application created from the Spring Archetype locally.~ → #1078

## Manual Checks?


docs-java/getting-started.mdx
```diff
@@ -159,7 +159,7 @@ Applications created with our SAP Cloud SDK Archetypes give you the possibility
 ```bash
 mvn clean install
 cd application/
-mvn spring-boot:run
+mvn spring-boot:run -D"spring-boot.run.profiles"=local
```
newtork, line 162: The local profile is not part of the archetype **yet** right?

Johannes-Schneider: No, it's not. 
However, I tested this change even without the new archetype and it still works (Spring will activate the profile, but since there is nothing configured for the profile it will also have no effect)

Johannes-Schneider: I extracted the change to #1078

MatKuhr (approved):

PR #1088, opened 2022-10-25
[Java] Update Overview Page

MatKuhr:
TBD

### DIFF
--- a/docs-java/overview.mdx
+++ b/docs-java/overview.mdx
@@ -18,6 +18,8 @@ keywords:
 
 import MvnBadge from '@site/src/sap/sdk-java/MvnBadge';
 import JavaFeatureTable from '@site/src/components/JavaFeatureTable';
+import useBaseUrl from '@docusaurus/useBaseUrl';
+import ThemedImage from '@theme/ThemedImage';
 
 <!-- vale on -->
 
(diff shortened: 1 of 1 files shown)


Johannes-Schneider:
I generally like the proposed content structure and also like the wording and abstraction level of the already changed documentation.

However, I think we should rework the _"Main Features Diagram"_:

image


* _"S/4 VDM"_: The term _'VDM"_ has not been established so far in this document. Instead, we are talking about _"(pre-generated) clients"_, which is what I would recommend to stick to, at least for this introduction. So changing the entry to something like _Pre-Generated S/4HANA Clients_ could make sense IMO. (Also: What's the difference between the S/4 VDM and our Convenience APIs?)
* _"Master Data Integration"_: MDI is still an internal service, which we never mentioned in the public docs portal. Therefore, I think we should remove that box from the diagram.
* _"CC, BL, Blockchain, ..."_: Is this really something we want to advertise on our Overview page?
* _"Deploy with Confidence"_: Same as for MDI.

MatKuhr:
The image is more of a placeholder, currently it is just a screenshot so I can't edit it :D 

At least CC must go out. Regarding internal stuff you are right, but maybe we want to reconsider / recheck what we can do here. It is an important part of the product and not having it well documented in one place is hurting visibility. But nothing to decide on this PR

docs-java/overview.mdx
```diff
@@ -18,19 +18,127 @@ keywords:
...
+### Availability and Licensing
+
+All libraries are available on [Maven Central](https://central.sonatype.dev/search?namespace=com.sap.cloud.sdk) under the [SAP Developer license](https://tools.eu1.hana.ondemand.com/developer-license-3_1.txt).
+The license of clients generated with any of the SAP Cloud SDK code generators depends on the used service spec license.
+
+### Release Schedule
+
+We usually release minor versions bi-weekly.
+
+We intend to release a new major version of the SAP Cloud SDK for JavaScript every twelve months.
```
cschubertcs, line 139: Is this now an active commitment? I would be happy =)

cschubertcs:
I like the direction this is going!

docs-java/overview.mdx
```diff
@@ -18,19 +18,147 @@ keywords:
...
```
CharlesDuboisSAP, line 159:
```suggestion
We intend to release a new major version of the SAP Cloud SDK for Java every twelve months.
```

docs-java/overview.mdx
```diff
@@ -18,19 +18,179 @@ keywords:
...
+
+You can then use these destinations for performing REST (OpenAPI or OData), RFC or SOAP requests.
+Or you can [get a ready-made Http Client](features/connectivity/sdk-connectivity-http-client#general-usage) for a destination to execute your own requests.
+
+### OData and OpenAPI
+
+The SAP Cloud SDK for Java also provides an easy and convenient type safe way to consume OData and OpenAPI services in your applications.
+<!-- vale off -->
+
+<ThemedImage
```
KavithaSiva, line 104: I am trying to use a plugin for trying to display gifs better. Will update the PR shortly if that works out.

docs-java/overview.mdx
```diff
@@ -18,19 +18,194 @@ keywords:
...
+  alt="Core packages"
+  sources={{
+    light: useBaseUrl('img/java-sdk-overview.png'),
+    dark: useBaseUrl('img/java-sdk-overview.png')
+  }}
+  className="center"
+  width="80%"
+/>
```
Johannes-Schneider, line 85: Although I like the gifs in general, currently this integration has the downside that **both** gifs will be transferred to the client on page load. 
This increases not only our page size but also loading times.

Additionally, users cannot stop the gif from playing, which might be annoying.

IMO, we shouldn't ship the gif integration as this since the UX is not good enough.
Instead, let's find a react component / Docusaurus Plugin to provides following features:

* Gif is **not** loaded on page load
* As a consequence, the gif is ofc also not played immediately. Instead, the customer should be able to toggle the gif playback.

PR #1089, opened 2022-10-27
Added instructions for fixing HttpClientFactory error

CharlesDuboisSAP:
## What Has Changed?

Added instructions in the FAQ to fix the [HttpClientFactory error](https://jira.tools.sap/browse/CLOUDECOSYSTEM-10259).
Added a link from the HttpClient page to the FAQ instructions.

## Manual Checks?

### DIFF
--- a/docs-java/faq.mdx
+++ b/docs-java/faq.mdx
@@ -92,24 +92,6 @@ java.util.ServiceConfigurationError: com.sap.cloud.security.client.HttpClientFac
 
 This issue is caused by a dependency scope conflict and can be solved by following [our guidance](./guides/manage-dependencies#avoiding-scope-conflicts).
 
-### I'm Observing a `More than 1 Custom HttpClientFactory service provider found. There should be only one` Exception
-
-An implementation of the `HttpClientFactory` is already provided by SAP Cloud SDK.
-
-1. Check if your implementation is needed.
-   Your application should be working without providing any `HttpClientFactory`.
-2. If you are implementing a missing feature, implement instead `com.sap.cloud.sdk.cloudplatform.connectivity.HttpClientFactory` and provide it to `HttpClientAccessor.setHttpClientFactory()`
-3. If the second step is not an option, the existing `HttpClientFactory` can be removed before the application starts:
-
-Place this static code in the application's main Class.
-
-```java
-static {
-    com.sap.cloud.security.client.HttpClientFactory.services.clear();
-    com.sap.cloud.security.client.HttpClientFactory.services.add(0, new MyCustomHttpClientFactory());
-}
-```
-
 ## OData Related Questions
 
 ### What Versions of OData Protocol Do You Support?
--- a/docs-java/faq.mdx
+++ b/docs-java/faq.mdx
@@ -94,8 +94,11 @@ This issue is caused by a dependency scope conflict and can be solved by followi
 
 ### I'm Observing a `More than 1 Custom HttpClientFactory service provider found. There should be only one` Exception
 
-After adding a `com.sap.cloud.security.client.HttpClientFactory`, there are now 2 HttpClientFactory.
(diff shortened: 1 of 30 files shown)


docs-java/features/connectivity/http-client.mdx
```diff
@@ -104,3 +104,8 @@ HttpClientAccessor.setHttpClientFactory(factory);
 ```
 
 From now on the custom factory will be used for every explicit and implicit HTTP request running through the SAP Cloud SDK.
+
+### Set up a `com.sap.cloud.security.client.HttpClientFactory`
```
Johannes-Schneider, line 108: I'm not sure whether we should mention this issue in our `HttpClient` documentation since the Security Lib `HttpClientFactory` has nothing to do with the SDK connectivity.
Therefore, I'd remove this section again.

docs-java/faq.mdx
```diff
@@ -92,6 +92,20 @@ java.util.ServiceConfigurationError: com.sap.cloud.security.client.HttpClientFac
 
 This issue is caused by a dependency scope conflict and can be solved by following [our guidance](./guides/manage-dependencies#avoiding-scope-conflicts).
 
+### I'm Observing a `More than 1 Custom HttpClientFactory service provider found. There should be only one` Exception
+
+After adding a `com.sap.cloud.security.client.HttpClientFactory`, there are now 2 HttpClientFactory.
```
Johannes-Schneider, line 97: I think we should add a little more context here.
For example, we could say that the Cloud SDK already brings an implementation of the mentioned interface, which should be working for most customers out of the box without any modification/action required from their side.

Therefore, our first advise for customers should be to check whether their custom implementation is **really** needed or whether they can leverage our `HttpClientAccessor.setHttpClientFactory` (which, under the hood, will also affect the Security Library) interface instead.

If that is not an option, only then should they apply the workaround.

CharlesDuboisSAP: You cannot provide a `com.sap.cloud.security.client.HttpClientFactory` in `HttpClientAccessor.setHttpClientFactory( com.sap.cloud.sdk.cloudplatform.connectivity.HttpClientFactory)
 `. I don't know if I can recommend this.

CharlesDuboisSAP: If the customers implement a missing feature with `com.sap.cloud.security.client.HttpClientFactory`. Can they provide this missing feature by providing an implementation of `com.sap.cloud.sdk.cloudplatform.connectivity.HttpClientFactory` to `HttpClientAccessor.setHttpClientFactory`?

docs-java/troubleshooting.mdx
```diff
@@ -231,3 +231,43 @@ You may experience the follow-up error message: `"ODataConnectionException: Conn
...
+```java
+@WebListener
+public class Config implements ServletContextListener {
+
+  public void contextInitialized(ServletContextEvent event) {
+    HttpClientFactory.services.clear();
+    HttpClientFactory.services.add(new MyHttpClientFactory());
+  }
+}
+
```
newtork, line 255:
```suggestion
```

docs-java/troubleshooting.mdx
```diff
@@ -231,3 +231,43 @@ You may experience the follow-up error message: `"ODataConnectionException: Conn
...
+  @Override
+  protected SpringApplicationBuilder configure(
+    final SpringApplicationBuilder application
+  ) {
+    HttpClientFactory.services.clear();
+    HttpClientFactory.services.add(new MyHttpClientFactory());
+  }
+}
+
```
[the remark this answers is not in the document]
CharlesDuboisSAP: This space is introduced by prettier, and is why the pipeline failed on the last commit

Johannes-Schneider (approved):
LGTM

PR #1113, opened 2022-11-07
[js] add a troubleshooting for missing `http-client` module

FrankEssenberger:
Relates to: https://github.com/SAP/cloud-sdk-backlog/issues/895

## What Has Changed?

Add a missing dependency section to the trouble shooting guide. @jjtang1985 it is mentioned in the ticket to add the guide to the site-map, which is here:

https://sap.github.io/cloud-sdk/sitemap.xml

generated from the `docusaurus.config.js`. I wonder if it is sufficient to have this in the general trouble shooting section? Should we add a dedicated page?


docs-js/troubleshooting.mdx
```diff
@@ -19,6 +19,22 @@ keywords:
 import useBaseUrl from '@docusaurus/useBaseUrl';
 import ThemedImage from '@theme/ThemedImage';
 
+## Module `@sap-cloud-sdk/http-client` not found
```
jjtang1985, line 22: The content looks good.
However, I could not find this by searching keywords in the algolia results.
Although this is not the focus of this ticket, this leads to a question to google SEO.
[pp] I would maybe use the original error message as the subtitle?

jjtang1985:
The sitemap also looks good.

docs-js/troubleshooting.mdx
FrankEssenberger: I trigger the algolia indexing.

FrankEssenberger: wait, I just did but this only makes sense after the merge I would say, right? I will put the error in the title perhaps this increases the results for the search engine as well

PR #1119, opened 2022-11-08
Update OData operations docs to include bound ops

fwilhe:
## What Has Changed?

Rewrite docs on OData operations to include bound functions/actions.

## Manual Checks?


docs-js/features/odata/common/function-imports/request-builder.mdx
```diff
@@ -1,19 +1,58 @@
...
+
+Unbound operations are generated into `actionImports` and `functionImports` respectively.
+
+The following code snippet shows the syntax for using them:
 
 ```ts
 const {
-  functionImports: { postGoodsIssue }
-} = warehouseOutboundDeliveryOrderService();
+  fooApi,
```
FrankEssenberger, line 24: [pp] perhaps a nicer name.

docs-js/features/odata/common/function-imports/request-builder.mdx
```diff
@@ -1,19 +1,58 @@
...
 To learn how to generate an API client from the service definition, look at [this article](features/odata/generate-odata-client.mdx).
 
-The example below creates a function import request builder for the service operation `PostGoodsIssue` and then executes it against your destination.
+### Using Operations
+
+Bound operations are generated into their respective entity class.
+
+Unbound operations are generated into `actionImports` and `functionImports` respectively.
+
+The following code snippet shows the syntax for using them:
```
FrankEssenberger, line 20: [pp] I would split the code snipest. One for unbound and one for bound.

docs-js/features/odata/common/function-imports/request-builder.mdx
```diff
@@ -1,19 +1,58 @@
...
 
 ```ts
 const {
-  functionImports: { postGoodsIssue }
-} = warehouseOutboundDeliveryOrderService();
+  fooApi,
+  actionImports: { unboundActionWithArguments, unboundActionWithoutArguments },
+  functionImports: {
```
FrankEssenberger, line 26: [req] I think this will appear for v2 and v4 now. Since the bound part is only available in v4 I think you need to split the documentation somehow and distribute it for the two cases.

FrankEssenberger: I think we should not invest to much to achieve this since we plan to restructure the docs anyway: https://github.com/SAP/cloud-sdk-backlog/issues/911

FrankEssenberger (commented):
LGTM - we really need to simplify the long doc for v2 and v4 in parallel. The whole reuse via the JS in mdx is also too complicated.

FrankEssenberger (approved):

PR #1121, opened 2022-11-10
[Java] Add gson dependency to OpenApi guide

cschubertcs:
## What Has Changed?

The code generated by the OpenApi generator actually also needs the `gson` dependency to compile.
This PR adds this to the Java OpenApi guide.

## Manual Checks?


newtork (changes requested):
We will cover this internally first.
We expect to mitigate the additional dependency in generated code.

Until we have final decision, I'd say let's keep the PR open.

KavithaSiva:
Closing as we decided to remove `gson` usages from the generated clients instead, refer to https://github.wdf.sap.corp/MA/sdk/pull/7813

PR #1138, opened 2022-11-21
Branch v3 js docs

fwilhe:
Branched v3 docs for js sdk

Default (for now) is v2. V3 docs have an 'unreleased' marker.


docusaurus.config.js
```diff
@@ -259,10 +259,17 @@ module.exports = {
         editUrl: 'https://github.com/SAP/cloud-sdk/edit/main',
         routeBasePath: 'docs/js',
         sidebarPath: require.resolve('./sidebarsDocsJs.js'),
-        lastVersion: 'current',
+        lastVersion: 'v2',
         versions: {
           current: {
+            label: 'v3',
+            banner: 'unreleased',
+            badge: false
```
deekshas8, line 267:
```suggestion
            badge: false,
            path: 'v3'
```

The paths have a `next` instead of `v3` (e.g. /cloud-sdk/docs/js/next/overview)

deekshas8 (changes requested):

docusaurus.config.js
```diff
@@ -259,10 +259,17 @@ module.exports = {
...
+          },
+          v2: {
             label: 'v2',
+            path: 'v2',
```
deekshas8, line 271:
```suggestion
```

[req] From my understanding (also described in the table [here](https://docusaurus.io/docs/versioning#overview)), the latest doc, which is v2 (as v3 is still wip), should still be accessible at `/cloud-sdk/docs/js/features/error-handling` and should not be changed to `/cloud-sdk/docs/js/v2/features/error-handling`. v3 will have the path `/cloud-sdk/docs/js/v3/features/error-handling`. Therefore, we should remove this line here and change back all paths which were changed by adding /v2/ (all redirects, navbar items)

When we finally want to make v3 the default, we just change the `lastVersion` to 'current' and add a `path` to v2 `path: 'v2'`

fwilhe: To explain what I did: I followed [option one outlined here](https://docusaurus.io/docs/versioning#configuring-versioning-behavior) because I wanted to use as much default behaviour of docusaurus as possible, and I think that this most closely resembles our situation.

fwilhe: Okay I think I got what you meant. I was under the impression that to do what you suggested we would have to tag a v3 now, but that's not the case. I've configured the v3 path for v3, reverted the changed links and v2 is now still available with the same links as before. Makes sense, thanks for the review.

FrankEssenberger:
Looks good overall with Deekshas suggestion all the links stay the same for v2 and work. I found one small thing. On main I still sse all the API references: 
Screenshot 2022-11-22 at 10 20 23
However on this branch I see only one:
Screenshot 2022-11-22 at 10 21 07
For v3 I do not see any references which is fine I think.

fwilhe:
@FrankEssenberger I think this is a unrelated change we discussed in a recent refinement. Not sure in which commit this was changed, but I'd be surprised if my PR caused that change. Also I think this is now the expected behaviour, so I don't know what to change.

notify @tomfrenken @marikaner @jjtang1985 

anyone diagrees?

deekshas8 (approved):
LGTM. I also do not see any difference in API Ref dropdown.
Main:
image

v3-docs branch:
image

PR #1139, opened 2022-11-21
DestinationProperty now public added

rnewbigging:
## What Has Changed?

Updated JavaDocs: Features -> Destinations to include explanation of using the now-public DestinationProperty class as predefined property keys for both creating and querying destination properties.

## Manual Checks?

- [ ] ~~Verify links still work, e.g., if `id` or the name of a file is changed.~~
- [ ] ~~Update the [feature matrix](https://github.com/SAP/cloud-sdk/blob/main/docs/components/data/features.js), e.g., if a feature is added.~~

### DIFF
--- a/docs-java/features/connectivity/destinations.mdx
+++ b/docs-java/features/connectivity/destinations.mdx
@@ -258,6 +258,23 @@ It can be a destination object that you defined in the Cloud Foundry Cockpit, or
 For that, call the `.property()` method on the destination.
 It expects a key-value pair with a key formatted as `URL.headers.<key>` and an arbitrary value string.
 
+Alternatively you can make use of a predefined `DestinationPropertyKey` available in the `DestinationProperty` class, for example:
+
+```java
+destination =
+            DefaultHttpDestination
+                .builder(server.baseUrl())
+                .property(DestinationProperty.URI, "value1")
+                .build();
+```
+
+These predefined keys can also be used when querying property values, for example:
+
+```java
+Option<String> uri =
+            destination.get(DestinationProperty.URI);
+```
+
 ## Provide Query Parameters for Destinations
 
 ### How It Works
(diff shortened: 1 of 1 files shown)


docs-java/features/connectivity/destinations.mdx
```diff
@@ -281,6 +298,23 @@ To include query parameters in the query string when calling to a target system
...
+                .property(DestinationProperty.URI, "value1")
+                .build();
+```
+
+These predefined keys can also be used when querying property values, for example:
+
+```java
+Option<String> uri =
+            destination.get(DestinationProperty.URI);
+```
```
Johannes-Schneider, line 316: This doesn't really fir the topic of "Adding Query Parameters".

docs-java/features/connectivity/destinations.mdx
```diff
@@ -258,6 +258,23 @@ It can be a destination object that you defined in the Cloud Foundry Cockpit, or
...
+
```
Johannes-Schneider, line 277: This doesn't really fit the topic of "Adding Headers via Destination Properties".

Johannes-Schneider (changes requested):
Personally, I think the added descriptions don't fit the chapters they were added to.

Instead, let's maybe introduce a new chapter where we describe how to access properties of a `Destination` instance, even without casting it first.

docs-java/features/connectivity/destinations.mdx
rnewbigging: Then I think I'm missing understanding of the purpose of the BLI as a whole. Was the point not to be able to reference/ create the properties without providing arbitrary strings? 

Should this then only be in a section about referencing properties?

Johannes-Schneider:
>> Personally, I think the added descriptions don't fit the chapters they were added to.
>>
>> Instead, let's maybe introduce a new chapter where we describe how to access properties of a Destination instance in a type-safe manner, even without casting it first.
>
> Do you think it would also be good to mention the other non-type-safe ways to access the properties? e.g. "Alternatively you can manage casting yourself with blah blah these methods?" Or do we only want to promote the type-safe feature?

No, I wouldn't mention that here I think.
With the shown API, customers should find the non-type-safe methods IMO.

Johannes-Schneider (approved):
LGTM

PR #1195, opened 2023-01-04
fix: v2 docs links

deekshas8:
## What Has Changed?

Closes https://github.com/SAP/cloud-sdk-backlog/issues/1034

Many links in v2 docs point to /v3/. This PR fixes them.
Updated the Styleguide to add information on path notation

## Manual Checks?


marikaner (changes requested):
Looks good, but we should add the same changes to v3, otherwise when we start working on v4 we will have the same issue.

deekshas8:
> but we should add the same changes to v3, otherwise when we start working on v4 we will have the same issue.

Yes, that's correct. I adjusted it for v3 as well. I also updated it to use the doc id instead of <filename>.mdx

FrankEssenberger (commented):
LGTM

FrankEssenberger (approved):
LGTM

PR #1209, opened 2023-01-12
[Java] Aligned slug names in v3 and v4

CharlesDuboisSAP:
## What Has Changed?

Page names have been changed in v4, making the dropdown to switch between v3 and v4 greyed out.

- v3 page names have been updated
- links were all updated and tested
- a broken link (localhost) was fixed

## Manual Checks?

- [x] Update the [feature matrix](https://github.com/SAP/cloud-sdk/blob/main/docs/components/data/features.js), e.g., if a feature is added.


docusaurus.config.js
```diff
@@ -387,62 +387,6 @@ module.exports = {
...
-            to: '/docs/java/features/bapi-and-rfc/overview'
-          },
-          {
-            from: '/docs/java/sdk-java-troubleshooting-frequent-problems',
-            to: '/docs/java/troubleshooting-frequent-problems'
-          },
-          {
-            from: '/docs/java/release-notes-sap-cloud-sdk-for-java',
-            to: '/docs/java/release-notes'
-          },
```
newtork, line 445: Why remove the lines here?

I would assume, this configures the auto-forwarding from one legacy URL to an updated URL path. If true, then I would assume, our SEO needs this to be enabled. Otherwise, google results will not pick up the URL changes. Please let me know, in case I got that wrong. :)

CharlesDuboisSAP: I think I removed this for the wrong reasons, the redirects were added for external links to keep working. It is still relevant.

docs-java_versioned_docs/version-v3/environments/sap-btp-kyma-environment.mdx
```diff
@@ -663,4 +664,4 @@ You can also refer to this additional material available for getting started wit
 
 - [Sample Application with Setup Guide](https://github.com/SAP-samples/kyma-runtime-extension-samples/tree/main/sample-cloudsdk-java)
 - [Blog Post on a Sample Use Case](https://blogs.sap.com/2022/01/11/use-sap-cloud-sdk-on-kyma-runtime-to-connect-to-external-systems-with-client-cert-authentication/)
-- [Using the Service Catalog with the SAP Cloud SDK](https://sap.github.io/cloud-sdk/docs/java/environments/sap-btp-kubernetes-environment-with-sap-gardener#excursion-bind-services-created-by-the-service-catalog)
```
Johannes-Schneider, line 666: I did a spot check with this link and found that there is apparently no redirect in place.
So when calling the old URL (as a customer with a bookmark might do), all I see is the 404 page.

I think we should fix this if the effort is not too high.

Johannes-Schneider (commented):
Updated slugs LGTM, however it seems we are missing some redirects. Not sure how much effort adding those would be, but I think it would be nice to have them.

docs-java_versioned_docs/version-v3/environments/sap-btp-kyma-environment.mdx
CharlesDuboisSAP: The old link works for me and the redirect from `sap-btp-kubernetes-environment-with-sap-gardener` to `kubernetes-gardener` is still in `docusaurus.config.js` line 395.

Johannes-Schneider: So when I check out your branch and run the project locally, using the old relative link `/cloud-sdk/docs/java/environments/sap-btp-kubernetes-environment-with-sap-gardener#excursion-bind-services-created-by-the-service-catalog` doesn't work.
(Of course, you must not use the entire old link, which was pointing to the live version of the docs prior to your change).

image

MatKuhr: IIRC redirects dont work when running `docusaurus start`. Try `docusaurus serve`. Not sure tho

Johannes-Schneider: That seems to work indeed! Thanks for sharing, I wasn't aware of that.

Johannes-Schneider (approved):
LGTM

PR #1225, opened 2023-01-19
[Java] Add section on Destination Caching improvement with optional user isolation

KavithaSiva:
## What Has Changed?

Adding a section on destination caching improvements done while caching destinations that do not require user propagation.

## Manual Checks?

- ~[x] Update the [feature matrix](https://github.com/SAP/cloud-sdk/blob/main/docs/components/data/features.js), e.g., if a feature is added.~


docs-java/features/connectivity/destinations.mdx
```diff
@@ -374,6 +376,28 @@ Therefore, we recommend adjusting the cache once at application startup.
...
+For the following authentication types:
+- `NoAuthentication`
+- `BasicAuthentication`
+- `ClientCertificateAuthentication`
+- `OAuth2ClientCredentials`
+- `OAuth2Password`
+
+destination caching is done per tenant.
+
+For these destinations,the improvement ensures that the cache is not filled up with multiple entries from the same tenant and different principals resulting in over 90% improvement in Cache hit ratio.
```
Johannes-Schneider, line 398: I'm not a fan of this style of documentation.
IMO, mentioning such improvements should be part of the release notes (which is already the case).
Therefore, describing how to further tweak the caching performance is a valuable hint IMO, but the rest can be omitted.

docs-java/features/connectivity/destinations.mdx
```diff
@@ -374,6 +376,30 @@ Therefore, we recommend adjusting the cache once at application startup.
...
+:::
```
Johannes-Schneider, line 401: I think such a hint should rather be part of the release notes (which it already is) as it refers to one specific change we introduced in one specific version.
Since our documentation is always for the latest available version, mentioning potentially old improvements is rather confusing IMO.

Instead, I'd say lets point the user to the token exchange strategies (especially the `EXCHANGE_ONLY` one) in case they are absolutely sure about the user token exchange and need even more performance.

docs-java/features/connectivity/destinations.mdx
```diff
@@ -374,6 +376,12 @@ Therefore, we recommend adjusting the cache once at application startup.
 
 :::
 
+:::note Hint
+
+For better performance, while querying for destinations **with user propagation**, users may provide the `EXCHANGE_ONLY` token exchange strategy.
+Refer [here](#token-exchange-options) for more details.
```
Johannes-Schneider, line 382:
```suggestion
Refer [here](#token-exchange-options) for more details.

```

CharlesDuboisSAP (approved):

PR #1226, opened 2023-01-20
chore: Only display latest API reference for the current major version

florian-richter:
## What Has Changed?

- There is only one API reference per major version
- API Reference is part of the versioned sidebar (and not in the header)
- Swizzled header dropdown removed (not needed anymore)
- LinkToLatestJsApiDocs replaced with markdown links (MANDATORY use `pathname:///` as prefix) -> Should link to relevant major version instead of latest!
- `cloud-sdk/api/latest` uses an HTML in the static folder to redirect instead of a docusaurus component -> Needs to be adjusted whenever a new major version is released
- Removed link to API References from release notes (as there are no more minor version specific api specs)

**Question for the review:**
> Make sure existing API docs links still work (redirect to most recent major version?)

-> This is currently not implemented. It would only be possible by a redirecting index.html for every version we previously had (assuming Github Pages is fixed... if we had our own hosting, we could fix it properly). The redirect plugin does not work for static content. Is this requirement really necessary?

## Manual Checks?


docs-js/features/connectivity/destination.mdx
```diff
@@ -103,7 +102,7 @@ Use [Register Destination](#register-destination) to cache destinations in produ
 :::
 
 The `destinations` environment variable is used for the destination lookup in the aforementioned order if they are presented.
-The value of the environment variable is expected to be a stringified JSON array, where the items adhere to the <LinkToLatestJsApiDocs slug="interfaces/sap_cloud_sdk_connectivity.Destination" name="Destination interface" />.
+The value of the environment variable is expected to be a stringified JSON array, where the items adhere to the [Destination interface](pathname:///api/v3/interfaces/sap_cloud_sdk_connectivity.Destination.html).
```
FrankEssenberger, line 105: [req] This seems dangerous. If we make a new version and copy the existings docs we will copy the version v3 to the new folder and reference the old docs. I guess this is not possible but to have a placeholder in the paht like `selected version in the drop down`.

FrankEssenberger: Or we have to add a step in the `generate new version docs` todo list that you have to do a `api/v3` search and replace.

FrankEssenberger: As discussed @florian-richter adds a todo to the docs.

FrankEssenberger (commented):
[q] Shouldn't there be a new folder `api/v1` `api/v2` and `api/v3` with the latest versions? Or do you want to generate them via the PR https://github.com/SAP/cloud-sdk-js/pull/3410/?

FrankEssenberger (approved):

PR #1231, opened 2023-01-24
chore: Swizzle NotFound page

florian-richter:
## What Has Changed?

Modified the not found page (404) to mention a few ways to resolve the missing links.

## Manual Checks?

### DIFF
--- a/src/theme/NotFound.js
+++ b/src/theme/NotFound.js
@@ -1,0 +1,35 @@
+import React from 'react';
+import Translate, { translate } from '@docusaurus/Translate';
+import { PageMetadata } from '@docusaurus/theme-common';
+import Layout from '@theme/Layout';
+export default function NotFound() {
+  return (
+    <>
+      <PageMetadata
+        title={translate({
+          id: 'theme.NotFound.title',
+          message: 'Page Not Found'
+        })}
+      />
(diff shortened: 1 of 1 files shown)


src/theme/NotFound.js
```diff
@@ -0,0 +1,35 @@
...
+            <div className="col col--6 col--offset-3">
+              <h1 className="hero__title">Page Not Found</h1>
+              <p>
+                The page you were looking for may have been moved or renamed.
+                Check that you are looking at the right language and the correct
+                version. Our search can also help.
+              </p>
+              <p>
+                If you were looking for the API reference of the SAP Cloud SDK
+                for JavaScript, you can find the latest version{' '}
```
tomfrenken, line 26: What's the purpose of the empty `{' '}`?

tomfrenken (approved):
LGTM, however since we include a JS-specific section, perhaps a Java-specific section should be added as well - consider syncing/informing our Java colleagues of this change.

I also left a small question.

src/theme/NotFound.js
florian-richter: This makes sure that there is a space between the last word and the link. HTML ignores spaces at the beginning and end of lines.

PR #1241, opened 2023-01-30
chore: Align generator options docs

deekshas8:
## What Has Changed?

Closes https://github.com/SAP/cloud-sdk-backlog/issues/972

- Add a "Default" column for OpenAPI
- Rename leftmost column to "Option"(remove "Alias") for both generators
- Order options similarly in the table

## Manual Checks?


docs-js/features/odata/generate-odata-client.mdx
```diff
@@ -97,22 +97,22 @@ All other files can be excluded from the generation - see the options below.
...
+| `--verbose`                                   |         `false`         |  By default, only errors, warnings and important info logs will be displayed. If set to true, all logs will be displayed.                                                                                                                                                                                                                                             |
+| `--skipValidation`                            |         `false`         |  Generation will stop if objects need renaming due to non-unique conditions or conflicts to JavaScript keywords. If you enable this option, conflicts are resolved by appending postfixes like '\_1'"                                                                                                                                                                 |
```
FrankEssenberger, line 115: [pp] I think this is not 100% true. We have protected keywords like `Entity` or `Service` and also uniqueness within the service, however the JS keywords like `const` are simply renamed not considering the option.

docs-js/features/openapi/generate-openapi-client.mdx
```diff
@@ -66,21 +66,21 @@ See [transpile](#Options) for more details in the options below.
...
+| `--skipValidation`                            |         `false`         | By default, the generation fails, when there are duplicate or invalid names for operations and/or path parameters after transforming them to camel case. Set this to true to enable unique and valid name generation. The names will then be generated by appending numbers and prepending prefixes.                                                              |
+| `-c`, `--config`                              |            -            | Set the path to the file containing the options for generation. If other flags are used, they overwrite the options set in the configuration file. If a directory is passed, a `config.json` file is read from this directory.                                                                                                                                    |
```
FrankEssenberger, line 83: [req] The `config` is present also for OData but I did not see it above.

docs-js/features/odata/generate-odata-client.mdx
deekshas8: I had just copied what we had originally, but I checked for JS keywords , and they are used without appending anything. 
I adjusted the documentation accordingly. Please have a look.

FrankEssenberger (approved):

PR #1250, opened 2023-02-02
chore: remove `VALUE_IS_UNDEFINED` from OData `servicePath` docs

deekshas8:
## What Has Changed?

Update servicePath docs after removing VALUE_IS_UNDEFINED

## Manual Checks?

### DIFF
--- a/docs-js/features/odata/generate-odata-client.mdx
+++ b/docs-js/features/odata/generate-odata-client.mdx
@@ -62,11 +62,12 @@ This file is used for customizing subdirectory naming and contains a mapping fro
 - `directoryName`: the name of the subdirectory the client code will be generated into.
 - `npmPackageName`: the name of the npm package, if a `package.json` file is generated.
   This information is optional.
-- `servicePath`: the URL path to be prepended before every request.
-  The url is read from the either the EDMX or Swagger file, if available.
-  Client generation will fail if a service path cannot be determined.
-  To prevent failure, set the `--skipValidation` option to true, in which case, `servicePath` value is set to `/` and a warn message is logged.
-  Or, specify the `servicePath` in the `options-per-service.json` file manually and re-generate the client.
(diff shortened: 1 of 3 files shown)


docs-js/features/odata/generate-odata-client.mdx
```diff
@@ -54,18 +54,19 @@ This will generate OData clients for all your service specifications.
...
 
 - `directoryName`: the name of the subdirectory the client code will be generated into.
-- `servicePath`: the URL path to be prepended before every request.
-  By default, this is read from the EDMX file, if available.
-  Otherwise, the value will be set as `VALUE_IS_UNDEFINED`, and an error will be logged.
-  In this case, you can specify `servicePath` in the `options-per-service.json` file manually and re-generate again.
 - `npmPackageName`: the name of the npm package, if a `package.json` file is generated.
   This information is optional.
+- `servicePath`: the URL path to be prepended before every request.
+  The url is read from the either the EDMX or Swagger file, if available.
```
FrankEssenberger, line 66:
```suggestion
- `servicePath`: Property to configure the  URL path manually if the automatic determination fails or does not give the desired value.
  The path is prepended before every request.
  If this option is not given the SDK tries to determine the value from either the EDMX or Swagger file.
```

FrankEssenberger: Also not super happy with my suggestion but I would like to strengthen the point that in most cases the automatic should be fine.

FrankEssenberger (approved):
LGTM - consider my suggestion.

fwilhe:
Hi @deekshas8,

we have [user feedback](https://github.com/SAP/cloud-sdk-js/issues/3506#issuecomment-1420549823) on the docs for value is undefined. Can you maybe check with the reporter of this issue if the new docs you're working on is better from their perspective?

Thanks

FrankEssenberger (approved):

PR #1265, opened 2023-02-09
[JS] Adjust broken dropdown

FrankEssenberger:
Relates to: https://github.com/SAP/cloud-sdk-backlog/issues/1047

## What Has Changed?

The documents links/ids for version 1 are aligned with version 2 so that the dropdown for the different versions can work.
On each page you can select version 1,2 or 3.
If the page does not exist on a version a little 🔝 icon is shown next to the version indicating that you will go to the overview page when you press the button.


fwilhe:
not sure how good the "top" icon is suited to communicate what we're doing here.. 

Screenshot 2023-02-21 at 09 45 34


But i think we can mere it, it is an improvement over what we have

fwilhe (approved):
not sure if it is perfect but it is an improvement over what we have so I'm okay with merging to get a better UX

PR #1269, opened 2023-02-13
[JS] Add setBasePath to the docs

tomfrenken:
## What Has Changed?

A previous pull request removed the docs relating to a renamed method.
This PR adds the docs back, now with the correct method

### DIFF
--- a/docs-js/features/odata/execute-odata-request.mdx
+++ b/docs-js/features/odata/execute-odata-request.mdx
@@ -228,10 +228,7 @@ It can also be adjusted per request by using the `setBasePath()` method:
 
 ```ts
 const { myEntityApi } = myEntityService();
-myEntityApi
-  .requestBuilder()
-  .getAll()
-  .setBasePath('my/custom/service/path');
+myEntityApi.requestBuilder().getAll().setBasePath('my/custom/service/path');
 ```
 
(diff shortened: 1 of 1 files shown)


docs-js/features/odata/execute-odata-request.mdx
```diff
@@ -224,6 +224,15 @@ If you want to set a query parameter in quotes (e.g. `language='en'`) you will h
 
 If a service specification contains a specification for the `basePath`, the SAP Cloud SDK generator generates an OData client with a default service path according to the specification (typically `'/sap/opu/odata/sap/'` for SAP S/4HANA services).
 When there is no such path defined in the specification, it can be manually set in the `options-per-service.json`.
+It can also be adjusted per request by using the `setBasePath()` method:
```
FrankEssenberger, line 227:
```suggestion
It can also be adjusted per request by using the `setBasePath()` method at runtime:
```

FrankEssenberger (approved):
[pp] I would link the section to the generator options on how to set the service path, because since version 3 there is so much logic to it:
- tries to determine it
- if not fails unless you set `skipValidation` the it is `/`
- or you do it manually with the service-per-options

So I would in general write here: look how you specify it during generation here. If you want to adjust it at runtime use:....

PR #1275, opened 2023-02-18
Suggest to add the import location for the filter functions

carlonnheim:
## What Has Changed?

It is not obvious for first-time consumers of this help page where the `and`, `or` etc. are to be imported from.


docs-js_versioned_docs/version-v2/features/odata/common/query-parameter/filter.mdx
```diff
@@ -10,6 +10,7 @@ Filters are built via the fields available on the schema property of the entitie
     - Or have first name 'Mallory'
 */
 const { businessPartnerApi } = businessPartnerService();
+const { or, and } = require('@sap-cloud-sdk/core');
```
marikaner, line 13: The core package only existed in version 1, for version 2 it should be loaded from the correct odata package (either odata-v2 or odata-v4). In this case it would be odata-v2. [=> docs](https://sap.github.io/cloud-sdk/api/v2/modules/sap_cloud_sdk_odata_v2.html#or)
```suggestion
const { or, and } = require('@sap-cloud-sdk/odata-v2');

```

marikaner (changes requested):
Once you sign the CLA we're good to go.

marikaner (approved):

marikaner:
@carlonnheim I am not allowed to merge your request as long as the CLA is not signed.

carlonnheim:
> @carlonnheim I am not allowed to merge your request as long as the CLA is not signed.

Sorry, missed your earlier message on this. I have done that now. Thanks!

PR #1341, opened 2023-04-12
[JS] chore: Remove removed generator options from js docs

tomfrenken:
## What Has Changed?

With this PR, the options which we removed or have hidden in version 3 are also removed from the usage example.


marikaner (approved):
Off with this! 👍

PR #1362, opened 2023-04-28
[Docu] New Destination-Type requires usage of typeguard

drvup:
Based on the changes of the `url` property on type `Destination` within SAP Cloud SDK v3, it's necessary to typecast `Destination` to be `HttpDestination`. The typeguard, SAP provides out of the box, is helping us to do so. Let's share this with the rest of the world :-)


tomfrenken:
Hey @drvup,

thanks a lot for your contribution already, somehow we missed that you now have to use a type guard or typecast the result.
We've discussed this in the team and will change the API so you won't have to use a type guard in the future, the change should come either in the next version or the version after that.

drvup:
Hi @tomfrenken ,

I really appreciate the feedback. It's fine for me to get rid again of the typeguard ;-) and I do see the reason why you did add this behavior. So for me, it's fine both ways. 

Kind regards,
Cedric

PR #1400, opened 2023-06-08
Update JCo deprecation warning

jjtang1985:
## What Has Changed?

Explain what you have changed.

## Manual Checks?

### DIFF
--- a/docs-java/features/bapi-and-rfc/overview.mdx
+++ b/docs-java/features/bapi-and-rfc/overview.mdx
@@ -18,8 +18,8 @@ import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
 :::danger Warning
-`RFC`, `JCo` and `SOAP` modules are **deprecated** and will be removed in version 5 of the SAP Cloud SDK.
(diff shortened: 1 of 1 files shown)


docs-java/features/bapi-and-rfc/overview.mdx
```diff
@@ -18,8 +18,8 @@ import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
 :::danger Warning
-`RFC`, `JCo` and `SOAP` modules are **deprecated** and will be removed in version 5 of the SAP Cloud SDK.
-After that, you can still use our SDK version 4, as it will be maintained for another half a year (until around Q1 2024).
+`RFC`, `JCo` and `SOAP` modules are **deprecated**.
+You can still use our SDK version 4, as it will be maintained for another half a year (until around Q1 2024).
```
MatKuhr, line 22: somewhat unfortunate situation: We probably don't want to write "it will not be removed in v5", because we don't want to promise that. So currently we kinda leave the reader guessing when it will be removed..

We could also write something like "the exact timeline of when these modules will be removed is not determined yet". A bit awkward, but a bit clearer...

MatKuhr (approved):

PR #1410, opened 2023-06-20
Update how-to-add-resilience.mdx

marikaner:
Using the resilience middleware in SAP Cloud SDK for JS was incorrectly documented, when using it with `executeHttpRequest()`.

Relates to https://github.com/SAP/cloud-sdk-js/issues/3941


marikaner:
`resilience()` DOES return an array. So this is wrong 🤦‍♀️

PR #1469, opened 2023-08-07
[Java] Reference jsr305 dependency in OData VDM projects

newtork:
## What Has Changed?

Mention JSR-305 dependency requirement.


CharlesDuboisSAP:
Should you add this to the v5 doc as well?

PR #1481, opened 2023-08-17
Add some description for the removed String Utils classes

cschubertcs:
## What Has Changed?

This PR adds some guidance regarding the removed String Utils classes that are removed in SDK v5.

### DIFF
--- a/docs-java_versioned_docs/version-v5/guides/5.0-upgrade.mdx
+++ b/docs-java_versioned_docs/version-v5/guides/5.0-upgrade.mdx
@@ -73,6 +73,7 @@ We will publish the full migration guide when we release v5.
 
 <details>
   <summary>Full list of deprecated API</summary>
+
   <table>
     <thead>
       <tr>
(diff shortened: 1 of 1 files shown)


docs-java_versioned_docs/version-v5/guides/5.0-upgrade.mdx
```diff
@@ -111,6 +112,44 @@ We will publish the full migration guide when we release v5.
...
+          Use libraries like{' '}
+          <a href="https://commons.apache.org/proper/commons-codec/">
+            Apache Commons Codec
+          </a>
+          , which offers a{' '}
+          <a href="https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base64.html">
+            Base64
+          </a>{' '}
+          class.
+        </td>
```
Johannes-Schneider, line 142: There is no need to use any "external" libraries for `Base64` since there is the `java.util.Base64` ([docs](https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html)) class ;)

cschubertcs: Fair, but I didn't see a 1:1 `isBase64` method in those, so I opted for the commons-codec library^^
But in the end we could also direct to the standard library one.

Johannes-Schneider: Good point 👍

Johannes-Schneider (approved):
LGTM

PR #1507, opened 2023-09-06
Add docs for removal of cli modules

cschubertcs:
## What Has Changed?

This PR adds some documentation on how to use our generators from the cli after we removed the cli modules.


Johannes-Schneider (approved):
LGTM

PR #1514, opened 2023-09-11
Java: `Destination` Migration Guide

Johannes-Schneider:
## What Has Changed?

This PR adds a description of `Destination` related changes to our v5 migration guide as well as the most important changes as part of our release notes.


docs-java_versioned_docs/version-v5/guides/5.0-upgrade.mdx
```diff
@@ -23,6 +23,45 @@ We will publish the full migration guide when we release v5.
 
 :::
 
+## Changes to the `Destination` API
```
MatKuhr, line 26: While I like having the motivations and backgrounds, I think for a migration guide this is a bit too much. In the guide I would only include minimal background information and rather list the specific actions developers should take to migrate (compare how our [V4 upgrade guide](https://sap.github.io/cloud-sdk/docs/java/guides/4.0-upgrade) is written).

We could keep this version but move it to a separate "more info / background" kinda document... or we make it a blog post or something like that..

docs-java_versioned_docs/version-v5/release-notes/release-notes-0-to-14.mdx
```diff
@@ -6,8 +6,18 @@
 
 ### Compatibility Notes
 
+- All `Destination` implementations have been removed with `DefaultHttpDestination` and `DefaultRfcDestination` being the only exceptions.
```
MatKuhr, line 9: I feel like this may be a bit too short / generic considering the level of detail of all other content that will go here

docs-java_versioned_docs/version-v5/guides/5.0-upgrade.mdx
Johannes-Schneider: I was thinking in that direction as well.
However, there is actually not too much developers need to do, right?

Basically just:

1. remove any calls to `.decorate(...)`
2. Make sure you are not declaring any concrete `Destination` types besides `DefaultHttp/RfcDestination` (i.e. `ScpCfHttpDestination` → `HttpDestination`)
3. Optional: Remove calls to `.asHttp/Rfc` and simply use `Destination` everywhere.
4. Done

Personally, I wasn't sure whether such a list is actually helpful?

docs-java_versioned_docs/version-v5/guides/5.0-upgrade.mdx
```diff
@@ -23,6 +23,108 @@ We will publish the full migration guide when we release v5.
...
+
+   // ******
+   // AFTER
+   // ******
+
+   Destination destination = DestinationAccessor.getDestination("my-destination");
+
+   try {
+       new MyODataService().getAllEntities().executeRequest(destination);
+   } catch (DestinationAccessException e) {
```
MatKuhr, line 75: I'm not sure this will be correct.. we would probably wrap the `DestinationAccessException` inside an `ODataRequestException`. Depends of course how we implement the change 😄

docs-java_versioned_docs/version-v5/guides/5.0-upgrade.mdx
```diff
@@ -23,6 +23,108 @@ We will publish the full migration guide when we release v5.
...
+2. (Optional) Remove any invocations of `Destination#asHttp` or `Destination#asRfc`.
+   A big part of the `Destination` cleanup was to straighten our public APIs that previously expected specific sub-types of `Destination` or `DestinationProperties` (e.g. `HttpClientAccessor#getHttpClient(HttpDestinationProperties)`).
+   These APIs now all simply rely on `Destination`, which makes casting via `#asHttp` or `#asRfc` unnecessary.
+
+3. Adjust error handling.
+   Prior to version 5 of the SAP Cloud SDK, `Destination#asHttp` would throw an Exception in case the Destination Service wasn't able to provide a valid `Authorization` header suggestions.
+   This has been changed so that `asHttp` will no longer throw an Exception.
+   Instead, calling the `getHeaders` method on an `HttpDestination` will throw an exception if any `Authorization` header suggestion provided by the Destination service contained an error.
+
+   This change is **very important** to keep in mind because it influences where applications are expected to catch Exceptions.
```
MatKuhr, line 45:
```suggestion
   This change is **important** to keep in mind because it influences where applications are expected to catch Exceptions.
```
Maybe just "important" is enough here. Also, if we think this is the most important change to adopt, maybe we should make it the first item of the list?

docs-java_versioned_docs/version-v5/guides/5.0-upgrade.mdx
```diff
@@ -23,6 +23,108 @@ We will publish the full migration guide when we release v5.
...
+   For example, code that looks like shown below should be updated:
+
+   ```java
+   // ******
+   // BEFORE
+   // ******
+
+   Destination destination = DestinationAccessor.getDestination("my-destination");
```
MatKuhr, line 53: This may indicate that `getDestination` doesn't throw. I get the intention of not having it together with `.asHttp()` in one try, otherwise it looks like there is no change 😄

docs-java_versioned_docs/version-v5/guides/5.0-upgrade.mdx
```diff
@@ -23,6 +23,108 @@ We will publish the full migration guide when we release v5.
...
+   try {
+       new MyODataService().getAllEntities().executeRequest(destination);
+   } catch (DestinationAccessException e) {
+       throw new IllegalStateException("Destination service wasn't able to provide a valid HTTP destination.", e);
+   } catch (ODataException e) {
+       throw new IllegalStateException("Unable to retrieve odata entities.", e);
+   }
+   ```
+
+4. Replace any mentions of `Destination` / `DestinationProperties` implementations other than `DefaultHttpDestination` and `DefaultRfcDestination`.
```
MatKuhr, line 82: To me this at first reads like "remove any usage of `Destination` and `DestinationProperties`". I think the full list of removed destination implementations is probably shorter than the explanation, I would just list them in a `details` block.

docs-java_versioned_docs/version-v5/guides/5.0-upgrade.mdx
```diff
@@ -23,6 +23,108 @@ We will publish the full migration guide when we release v5.
...
+   Version 5 of the SAP Cloud SDK removes all other implementations of `Destination` and `DestinationProperties` to unify the APIs and implementations.
+   As a consequence, classes such as `ScpCfHttpDestination`, `ErpHttpDestination`, and `WrapperDestination` (see the table further down for all removed classes) have been removed.
```
MatKuhr, line 84:
```suggestion
   As a consequence, classes such as `ScpCfHttpDestination`, `ErpHttpDestination`, and `WrappedDestination` (see the table further down for all removed classes) have been removed.
```

docs-java_versioned_docs/version-v5/guides/5.0-upgrade.mdx
Johannes-Schneider: Yeah, I was also wondering... I guess we have to update this part of the guide anyways, so I thought let's keep it in sync with the actual behavior.

docs-java_versioned_docs/version-v5/guides/5.0-upgrade.mdx
Johannes-Schneider: > maybe we should make it the first item of the list?

I was also considering doing that. However, I felt like this step might make more sense after users saw the change about the `#asHttp` being optional?

docs-java_versioned_docs/version-v5/guides/5.0-upgrade.mdx
Johannes-Schneider: I tried making the changes more apparent by using a `diff` block.

docs-java_versioned_docs/version-v5/guides/5.0-upgrade.mdx
Johannes-Schneider: I re-phrased this bullet point to simply point to the list of removed APIs (which contains all removed classes and their replacement, if any).

MatKuhr (approved):

PR #1526, opened 2023-09-14
[java] Switch support channel to it's own repository

jjtang1985:
## What Has Changed?

Explain what you are changing and why, if it isn't obvious from the diff.

**If you are working on the Java v4 docs, remember to also apply the changes to v5**

PR #1543, opened 2023-09-28
[java] Update support board link for overview page

jjtang1985:
## What Has Changed?

In the previous [PR](#1526), we updated the support board link.
We should do the same for the overview page.


CharlesDuboisSAP (changes requested):
There are 7 more links to update:
image

CharlesDuboisSAP (approved):

PR #1557, opened 2023-10-10
Fixed crash when inspecting any page

CharlesDuboisSAP:
## What Has Changed?

Right click -> Inspect any page of the docs would crash it.
Fixed the JavaScript code that would crash


FrankEssenberger (approved):
LGTM you could also use conditional chaining in such cases: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

PR #1559, opened 2023-10-11
[V5] framework module removed

rnewbigging:
## What Has Changed?

Added line to migration guide about `frameworks` removal and child module movement. Updated dependency ids.


jjtang1985 (approved):
LGTM

Issue #1578, opened 2023-10-24
Webpack bundling failure for browser because of "node:url" is not handled by plugins


magician-margatroid:
Hi, according to config in doc [Use the SAP Cloud SDK in the Browser](https://sap.github.io/cloud-sdk/docs/js/guides/browser), at present, the bundling in my environment will fail with this error:
```bash
Module build failed: UnhandledSchemeError: Reading from "node:url" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
```

The webapp to reproduce this phenomenon: [cloud-sdk-in-react](https://github.com/magician-margatroid/cloud-sdk-in-react) with steps listed on its [README.md](https://github.com/magician-margatroid/cloud-sdk-in-react/blob/master/README.md).

Taking the advantage of [normal-module-replacement-plugin](https://webpack.js.org/plugins/normal-module-replacement-plugin/) could be a possible solution dealing with this issue.

tomfrenken:
Hey @magician-margatroid,

This issue was introduced when we started prefixing our node dependencies with `node:` as this has recently become a best practice, however, we failed to realize this might cause issues in the browser.

Thank you very much for your in-depth investigation as well as your proposed solution, we will adjust our documentation.

magician-margatroid:
> Hey @magician-margatroid,
> 
> This issue was introduced when we started prefixing our node dependencies with `node:` as this has recently become a best practice, however, we failed to realize this might cause issues in the browser.
> 
> Thank you very much for your in-depth investigation as well as your proposed solution, we will adjust our documentation.

Hi @tomfrenken thx a lot and it's my pleasure.

PR #1579, opened 2023-10-24
fix: use NormalModuleReplacementPlugin

magician-margatroid:
Use NormalModuleReplacementPlugin for internal node modules with node-protocol when compiling for browser.

## What Has Changed?

Applying changes into [Use the SAP Cloud SDK in the Browser](https://sap.github.io/cloud-sdk/docs/js/guides/browser) in current and v2 docs of JavaScript, to include a possible solution for [#1578](https://github.com/SAP/cloud-sdk/issues/1578) where `node:` protocol will fail for bundling towards brower.


tomfrenken (approved):
LGTM

PR #1581, opened 2023-10-27
[Java] Add release notes for 4.26.0

cschubertcs:
This adds the release notes for the latest SDK v4 release.


Johannes-Schneider (approved):
Are all of these dependencies part of our BOM? (Because usually those are the only ones we are mentioning in our release notes)

PR #1621, opened 2023-12-05
Java SDK v5 release notes

cschubertcs:
## What Has Changed?

Taken over and filtered/adapted from the release notes in the v5 branch.


MatKuhr:
@cschubertcs FYI I did some significant updates in [this commit](https://github.com/SAP/cloud-sdk/pull/1621/commits/b3b85069a6270cb5f96e722f67daced78a7aa4b8).

Merging the current state, although still a bit WiP, so that we have a baseline on which we can switch the active (i.e. default) version to v5 😉

MatKuhr (approved):

PR #1639, opened 2023-12-13
[Java] Added replacement for `getVcapApplication`

CharlesDuboisSAP:
## What Has Changed?

The service binding lib doesn't replace `getVcapApplication`, we need to add a code snippet to help customers migrate

PR #1641, opened 2023-12-14
suggestion for #1639

Johannes-Schneider:


docs-java/guides/5.0-upgrade-steps.mdx
```diff
@@ -640,29 +640,31 @@ Once the build doesn't complain about missing dependency versions anymore, you c
       <tr>
         <td>CloudPlatform</td>
         <td>-</td>
-        <td>No replacement, as all platform specifics are handled by the <a href="https://github.com/SAP/btp-environment-variable-access">Service Binding library</a>.</td>
+        <td>No SDK replacement. <br />
+        Handling bound applications (i.e. <code>VCAP_SERVICES</code>) is done via the <a href="https://github.com/SAP/btp-environment-variable-access">Service Binding library</a>.<br />
+        Handling application information (i.e. <code>VCAP_APPLICATION</code>) can be done manually: <br />
+        <pre>new Gson().fromJson(JsonSanitizer.sanitize(System.getenv("VCAP_APPLICATION")), new TypeToken &lt;Map&lt;String, JsonElement>>(){}.getType());</pre>
```
newtork, line 646: (Minor)

Do you think it's necessary to sanitize? (It means another dependency to the user) 🤔 
```suggestion
        <pre>new Gson().fromJson(System.getenv("VCAP_APPLICATION"), new TypeToken &lt;Map&lt;String, JsonElement>>(){}.getType());</pre>
```

The current code could be even further simplified with...
```suggestion
        <pre>new Gson().fromJson(environmentVariable, JsonObject.class).asMap();</pre>
```

newtork (approved):
Suggestion is not a requirement for merging.
LGTM

docs-java/guides/5.0-upgrade-steps.mdx
Johannes-Schneider, line 646: @newtork 
I fear that some consumers might be confused by the `environmentVariable` (where does it come from...?).

So what do you think about this:

```suggestion
        <pre>new Gson().fromJson(System.getenv("VCAP_APPLICATION"), JsonObject.class).asMap();</pre>
```

CharlesDuboisSAP (approved):

PR #1673, opened 2024-01-15
fix: Build Documentation Workflow

Johannes-Schneider:
## What Has Changed?

This PR fixes our `Build Cloud SDK Documentation` workflow, which is currently failing due to issues when checking PRs using `vale`.


.github/workflows/build_documentation.yml
```diff
@@ -10,21 +10,26 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      - name: Determine Changed Files
+        id: changed-files
+        run: |
+          ALL_FILES=$(gh pr view ${{ github.event.pull_request.number }} --json files --jq '.files.[] | select(.additions > 0) | .path')
```
Johannes-Schneider, line 16: **Note for reviewers**:

We are checking files that contain _at least one addition **only**_.

One the one hand, this drastically reduces the amount of time required for `vale` to run.
On the other hand, this also avoids "false positives" (i.e. reporting those findings that are already present on our `main` branch).

CharlesDuboisSAP (commented):
[fail_on_error is set to false by default.](https://github.com/errata-ai/vale-action?tab=readme-ov-file#fail_on_error-default-false). Was this intended?

Johannes-Schneider:
@CharlesDuboisSAP wrote:

> [fail_on_error is set to false by default.](https://github.com/errata-ai/vale-action?tab=readme-ov-file#fail_on_error-default-false). Was this intended?

I guess it never was set to `true`, but I agree: It should be set to `true`. Good catch!

CharlesDuboisSAP (approved):

PR #1681, opened 2024-01-19
Java: Document Http Client Cache

MatKuhr:


styles/SAP/Headings.yml
```diff
@@ -1,7 +1,7 @@
 extends: capitalization
 message: "'%s' should use title-style capitalization."
 link: 'https://capitalizemytitle.com/style/AP/#'
-level: warning
+level: suggestion
```
MatKuhr, line 4: Vale is just drunk, the capitalization check doesn't work properly with the GitHub action. Tried fixing it by e.g. updating the vale version, didn't work. So effectively disabling the check.

styles/SAP/SentenceLength.yml
```diff
@@ -1,6 +1,6 @@
 extends: occurrence
-message: 'Try to keep sentences short (< 30 words).'
+message: 'Try to keep sentences short (< 45 words).'
 scope: sentence
 level: warning
-max: 30
+max: 45
```
MatKuhr, line 5: 30 is too restrictive IMO. Making the check more lenient as to make PR builds more robust, otherwise one has to edit a lot of lines unrelated to the PR itself.

Johannes-Schneider (approved):
LGTM

PR #1685, opened 2024-01-22
chore: Add docs on mail-client debug options

deekshas8:
## What Has Changed?


mr-flannery (approved):
very nice, thank you!

PR #1691, opened 2024-01-25
Update Docs for Transparent Proxy 1.4

MatKuhr:
Refer to the [release notes](https://help.sap.com/whats-new/cf0cb2cb149647329b5d02aa96303f56?Component=Connectivity&locale=en-US).


docs-java/environments/gardener.mdx
```diff
@@ -467,30 +467,25 @@ Create the following YAML file:
...
-    name: 'example-dest-onprem' # name from BTP Cockpit
+    name: '*'
   destinationServiceInstanceName: dest-service-instance-example # can be ommited if config.destinationService.defaultInstanceName was provided to the helm chart of the transparent proxy upon installation
 ```
 
-- Adjust the name `example-dest-onprem` to the On-Premise destination you defined in your SAP BTP Cockpit.
-
-It will create a Kubernetes resource in your cluster, pointing to the destination you created in the cockpit.
+It will create a Kubernetes resource in your cluster, allowing you to create requests against arbitrary destinations.
 Apply the YAML with `kubectl apply` to take effect.
```
MatKuhr, line 478: Should we mention here that it has to be created in the namespace of the transparent proxy?

docs-java/environments/kyma.mdx
```diff
@@ -568,8 +568,6 @@ Please note that the current implementation of the Transparent Proxy does not ye
   <summary>Constraints when using the Transparent Proxy</summary>
 
 - Private Link not yet supported
```
MatKuhr, line 570: Is the statement about private link still correct?

docs-java/environments/kyma.mdx
```diff
@@ -620,16 +656,50 @@ List<SalesArea> execute = new DefaultSalesAreaService().getAllSalesArea() // exa
...
+
+To make troubleshooting easier the Transparent Proxy adds additional response headers to provide more information about where an error occurred.
+For the above example of executing OData requests you can access the response headers as follows:
+
+```java
+try {
+    // execute OData request
+} catch (ODataResponseException e) {
+    System.out.println(e.getHttpCode());
+    // the Transparent Proxy will attach additional response headers in case an error occurred
```
MatKuhr, line 702: can we add the specific header names to look out for?

docs-java/environments/kyma.mdx
```diff
@@ -580,36 +582,71 @@ The official [documentation of the Transparent Proxy](https://help.sap.com/docs/
 
 ### Create a Kubernetes Resource
 
+You can either configure connectivity to individual destinations, or for arbitrary destinations in your destination service instance or subaccount (dynamic destinations).
+
+<Tabs 
+  groupId="dynamic-dest"
+  defaultValue="single"
```
Johannes-Schneider, line 589: Why are we not making the `dynamic` approach the default case? It sounds like this is way easier to maintain.

Johannes-Schneider (commented):
Generally LGTM - just one question left from my side.

docs-java/environments/kyma.mdx
```diff
@@ -580,36 +582,71 @@ The official [documentation of the Transparent Proxy](https://help.sap.com/docs/
...
+  values={[
+    {label: "Individual Destination", value: "single"},
+    {label: "Dynamic Desitnations", value: "dynamic"}
+  ]}>
+<TabItem value="single">
+
 ```java
 DefaultHttpDestination destination = DefaultHttpDestination
-        .builder("http://example-dest.namespace/")
+        .builder("http://my-destination.namespace/")
```
newtork, line 649: (Question)

Is `namespace` a placeholder or the real "top-level domain"?

docs-java/environments/kyma.mdx
MatKuhr: I considered this too and for me two reasons stood out: complexity and security. We have seen in the past (with megaclite) that people tend to struggle with understanding the concept and I figured the slightly simpler case make that a bit easier. And in general it is typically better to put the simple case first. 

For security, remember that with dynamic destinations we are exposing all destinations (incl. the subscriber destinations) in the K8S cluster (in Kyma in the Istio mesh) without any (additional) authorization required. Any pod can access them at will, which means one has to take extra caution compared to CF, where only apps with destination service bindings + the necessary auth flow can use the destinations.

So if you only have 1 or 2 destinations you care about, creating individual K8S resources for them might just be better.

docs-java/environments/kyma.mdx
```diff
@@ -620,16 +657,60 @@ List<SalesArea> execute = new DefaultSalesAreaService().getAllSalesArea() // exa
...
+</TabItem>
+<TabItem value="dynamic">
 
-This will target the Kubernetes resource `example-dest` we created above which will be resolved to the destination we configured in that resource.
-The code above shows an example how you can use it to perform an OData request.
+```java
+DefaultHttpDestination destination = DefaultHttpDestination
+        .builder("http://dynamic-destination.namespace/")
+        // use the name of the desitnation you configured in the BTP Cockpit
+         .header(new Header("X-Destination-Name", "my-destination"))
```
jjtang1985, line 667: question:
so the difference is that:
- single destination has hard coded config for "my-destination"
- dynamic destination enables the programatically way in the application

jjtang1985 (commented):
LGTM.
Only some minor comments/questions.

docs-java/environments/gardener.mdx
```diff
@@ -443,77 +446,8 @@ We generally recommend to go with approach (1) unless the given constraints appl
...
-:::note
-Repeat this process for all _destination names_ you want to use in your application.
-:::
-
-:::tip Connecting to Cloud systems
-The above approach is not limited to destinations of proxy type `ON_PREMISE`.
-`INTERNET` destinations are equally supported.
-:::
+Using the SAP Cloud SDK with the Transparent Proxy is identical to how it is used on Kyma.
+Please refer to [this documentation](/docs/java/environments/kubernetes-kyma#on-premise-connectivity).
```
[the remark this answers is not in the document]
MatKuhr: typically I use absolute because if the text ever get's moved elsewhere it would still work, with relative links it's then hard to understand what the link was supposed to point at..

docs-java/environments/kyma.mdx
MatKuhr: correct 👍🏻

docs-java/environments/kyma.mdx
```diff
@@ -580,36 +582,71 @@ The official [documentation of the Transparent Proxy](https://help.sap.com/docs/
...
+  name: dynamic-destination
 spec:
   destinationRef:
-    name: 'example-dest-onprem' # name from BTP Cockpit
+    name: '*'
   destinationServiceInstanceName: dest-service-instance-example # can be ommited if config.destinationService.defaultInstanceName was provided to the helm chart of the transparent proxy upon installation
 ```
 
-- Adjust the name `example-dest-onprem` to the On-Premise destination you defined in your SAP BTP Cockpit.
+This will create a Kubernetes resource in your cluster, allowing you to create requests against arbitrary destinations.
```
Johannes-Schneider, line 627: Should we point out that this approach comes with the "risk" of having all (i.e. for all tenants) destinations accessible by any application within the cluster?

MatKuhr: Done 👍🏻

PR #1701, opened 2024-02-07
refactor: Remove Mentions of `Zone` Concept

Johannes-Schneider:
## What Has Changed?

This PR refactors our documentation about multitenancy so that it no longer mentions the `Zone` concept.


cschubertcs (commented):
Should this also be updated in version 3/4 of the documentation?

Johannes-Schneider:
@cschubertcs wrote:

> Should this also be updated in version 3/4 of the documentation?

I wouldn't touch any V3 documentation.
For V4, I thought about it, but as we are (most probably) not back-porting our JavaDocs updates that we are currently doing in V5, we would end up in a somewhat inconsistent state.
So I didn't do it.

newtork (approved):

PR #1724, opened 2024-02-27
Update generate-rest-client.mdx

newtork:
## What Has Changed?

### DIFF
--- a/docs-java/features/rest/generate-rest-client.mdx
+++ b/docs-java/features/rest/generate-rest-client.mdx
@@ -41,6 +41,33 @@ The generated Java classes need the following dependency to be present in your p
 
 To use the generated client library, either the SAP Cloud SDK BOM should be part of your project's `<dependencyManagement>` section, or the version for the `openapi-core` artifact must be mentioned explicitly here.
 
+## Generated Java Classes and Usage
+
+The following chapter will explain how to use the OpenAPI Generator Maven Plugin.
+A sample result can be seen in our [open-source public repository](https://github.com/SAP/cloud-sdk-java/tree/main/datamodel/openapi/openapi-api-sample): the soda store example.
+
+After generation the following API is available:
+```java
+Destination destination;
+
+// create the service with a destination
+SodasApi service = new SodasApi(destination);
+
+// create an entity object
+SodaWithId soda = new SodaWithId().id(0L).name("Cola").brand("SAP-Cola").quantity(100).price(1.5f);
+
+// execute request for entity update
+service.sodasPut(soda);
+```
+
+Services can be instantiated with:
+```
+new SodasApi( Destination )
+new SodasApi( new ApiClient() )
+new SodasApi( new ApiClient(RestTemplate) )
+new SodasApi( new ApiClient(Destination) )
+```
+
 ## Customize Java Class and Method Names with OpenAPI Vendor Extensions
 
 The OpenAPI generator uses the `tag` field to determine the Java service class name and the `operationId` to determine the Java method for each service operation.


docs-java/features/rest/generate-rest-client.mdx
```diff
@@ -41,6 +41,33 @@ The generated Java classes need the following dependency to be present in your p
 
 To use the generated client library, either the SAP Cloud SDK BOM should be part of your project's `<dependencyManagement>` section, or the version for the `openapi-core` artifact must be mentioned explicitly here.
 
+## Generated Java Classes and Usage
+
+The following chapter will explain how to use the OpenAPI Generator Maven Plugin.
```
jjtang1985, line 46:
```suggestion
In this section, we will explain how to use the OpenAPI Generator Maven Plugin.
```
chapter vs section 😄

docs-java/features/rest/generate-rest-client.mdx
```diff
@@ -41,6 +41,33 @@ The generated Java classes need the following dependency to be present in your p
...
+A sample result can be seen in our [open-source public repository](https://github.com/SAP/cloud-sdk-java/tree/main/datamodel/openapi/openapi-api-sample): the soda store example.
```
jjtang1985, line 47: Great reference to our sample.

Question:
Would it be better to share the directory structure so we know the [api dir](https://github.com/SAP/cloud-sdk-java/tree/main/datamodel/openapi/openapi-api-sample/src/main/java/com/sap/cloud/sdk/datamodel/openapi/sample/api) is the location for all the APIs?
With that, it might be smooth to say there is a `SodaApi`, that we will use later as the starting point?

jjtang1985 (commented):
Look good in general.
I left some suggestions.

docs-java/features/rest/generate-rest-client.mdx
```diff
@@ -41,6 +41,50 @@ The generated Java classes need the following dependency to be present in your p
...
+service.sodasPut(soda);
+```
+
+Services can be instantiated in different ways:
+
+```java
+new SodasApi( Destination )
+new SodasApi( new ApiClient() )
+new SodasApi( new ApiClient(RestTemplate) )
+new SodasApi( new ApiClient(Destination) )
```
jjtang1985, line 85: As a user, I might have the following question:
Do we have some documentation for the different use cases mentioned here?
What would be the recommended way?
Is there any differences between `new SodasApi( Destination )` and `new SodasApi( new ApiClient(Destination) )`?
If we use the `new SodasApi( new ApiClient() )` or `new SodasApi( new ApiClient(RestTemplate) )` without a destination, can I inject the destination later or how can I tell the URL of the target system?

newtork: > Do we have some documentation for the different use cases mentioned here?

I added a few sentences to your questions. I would expect the JavaDoc to explain the fine details. _(I added a link to the branch)_

> What would be the recommended way?

Only using `Destination` of course. But I didn't feel comfortable hiding the features of `ApiClient`.

> Is there any differences between `new SodasApi( Destination )` and `new SodasApi( new ApiClient(Destination) )`?

The latter allows for customization of base-path, default headers and user-agent, via `ApiClient` reference. _(I added this to the branch)_

> If we use the `new SodasApi( new ApiClient() )` or `new SodasApi( new ApiClient(RestTemplate) )` without a destination, can I inject the destination later

You can't assign a destination after `ApiClient` is instantiated. (I see no reason in documenting what's not possible.)

> or how can I tell the URL of the target system?

Yes, `ApiClient` allows for setting/changing the URL via `setBasePath(String)`. (I think by linking the JavaDoc I don't need to explain this explicitly)_

jjtang1985 (approved):

PR #1728, opened 2024-02-29
fix: HTTP Client 5 Link

Johannes-Schneider:
Fixes SAP/cloud-sdk-java/issues/336

PR #1729, opened 2024-02-29
fix: replace maven central search links for our modules

Johannes-Schneider:
## What Has Changed?

Follow-up of #1728 

This PR replaces links that pointed to the old Sonatype UI for our modules.


cschubertcs (commented):
I found the following files that also had some `search.maven.org` references and seem v5 related (v3 and v4 have a lot more findings, but I see no need to update those as well:

- `docusaurus.config.js`
- `sidebarsDocsJava.js`
- `src/sap/sdk-data/data.js`

Not sure, whether they need to be changed, but might be a good idea to do it.

cschubertcs (approved):
lgtm

PR #1742, opened 2024-03-14
docs: CSRF Token Handling in Multi-Threaded Workloads

Johannes-Schneider:
## What Has Changed?

This PR adds a new section to both our OData v2 and v4 docs about how to correctly handle CSRF tokens when executing requests in parallel.
Furthermore, the section also summarizes our current knowledge about why this is needed in the first place.


cschubertcs (approved):
Sounds good to me!
I even learned, that it's okay to re-use CSRF token in a "session", I was under the impression it should be rotated per request.

PR #1748, opened 2024-03-19
fix: typo in generator doc

BenjYue:
## What Has Changed?

Found some minor typos in the generator docs. It should be `com.sap.cloud.sdk.datamodel` instead of `om.sap.cloud.datamodel` to work correctly.


MatKuhr (approved):
Thanks for fixing this!

PR #1850, opened 2024-07-02
chore: Add JS Kyma On-Premise documentation

tomfrenken:
## What Has Changed?

With this PR, we re-use the Java documentation for Kyma On-Premise connectivity, adjusted for JavaScript use.


docs-js/environments/migrate-sdk-application-from-btp-cf-to-kyma.mdx
```diff
@@ -217,3 +219,201 @@ Verify the resources and choose the host link to open the application in your br
...
+  ]}>
+<TabItem value="single">
+
+```ts
+import { registerDestination, getTenantId } from '@sap-cloud-sdk/connectivity';
+
+const destination = {
+  name: 'registered-destination',
+  url: 'http://my-destination.namespace/'
+  // for principal propagation make sure to set the auth type to set the forwardAuthToken flag to true:
```
MatKuhr, line 345:
```suggestion
  // for principal propagation make sure to set the set the forwardAuthToken flag to true:
```
I guess? 😄

MatKuhr: Also, where does the JS SDK pull the subscriber token from in this case? would you have to pass it as argument to `.execute(..)`? If so, maybe extend the example..

MatKuhr (commented):
instead of copy/pasting the Java docs, we could also only link to them, and only have the `Executing Requests` section. No strong opinion, but chances are we will forget to update both documents when updating either.

Finally, what about the gardener guide?

docs-js/environments/migrate-sdk-application-from-btp-cf-to-kyma.mdx
tomfrenken: Good point! 💀 that's why you're the architect

MatKuhr (approved):

PR #1907, opened 2024-09-12
Fix cf ssh command parameters

CharlesDuboisSAP:
## What Has Changed?

Replaced duplicate `onpremise_proxy_port` with new `onpremise_proxy_http_port`

### DIFF
--- a/docs-java/features/connectivity/006-running-locally.mdx
+++ b/docs-java/features/connectivity/006-running-locally.mdx
@@ -151,11 +151,12 @@ cf restart app-name
 
   - `credentials.onpremise_proxy_port` (we will refer to the value as _proxy-port_ hereafter)
   - `credentials.onpremise_proxy_host` (we will refer to the value as _proxy-host_ hereafter)
+  - `credentials.onpremise_proxy_http_port` (we will refer to the value as _proxy_http_port_ hereafter)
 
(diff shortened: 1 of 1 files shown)


MatKuhr (commented):
[both should work equally](https://github.com/SAP/cloud-sdk-java/blob/main/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java#L146), there should be no need to set both?

CharlesDuboisSAP:
> [both should work equally](https://github.com/SAP/cloud-sdk-java/blob/main/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java#L146), there should be no need to set both?

They represent different values. The first one is the local port, the second one is the remote port

MatKuhr:
> > [both should work equally](https://github.com/SAP/cloud-sdk-java/blob/main/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java#L146), there should be no need to set both?
> 
> They represent different values. The first one is the local port, the second one is the remote port

no, both represent the remote port. The service binding only declares remote ports, no local ports.

PR #1967, opened 2024-12-17
add top(1) to call to bussinessPartnerAPI

DenisDuev:
without it I get: "statusCode":500,"message":"Failed to get business partners - get request to https://sandbox.api.sap.com/s4hanacloud/sap/opu/odata/sap/API_BUSINESS_PARTNER failed! "}

Also noted here: https://github.com/SAP/cloud-sdk-js/issues/2896

## What Has Changed?

get the top(1) result because it shows an error


ZhongpinWang (approved):
Hi @DenisDuev , 

Thanks for your contribution. Although it just happens in the sandbox system, I think it won't hurt if we use the `top()` method in between to make it work.

PR #2019, opened 2025-02-20
Improve JS `getDestinationFromServiceBinding()` doc

ZhongpinWang:
## What Has Changed?

https://github.com/SAP/cloud-sdk-js/issues/5517

The current usage for using the `serviceBindingTransformFn` function is wrong. Also improved the overall doc which was confusing to the user.


tomfrenken (approved):
LGTM

PR #2023, opened 2025-02-21
[js] Update feature scope

jjtang1985:
## What Has Changed?


tomfrenken (approved):
LGTM

PR #2044, opened 2025-03-11
Remove node forge

CharlesDuboisSAP:
## What Has Changed?

Removed the Node Forge dependency that has a License risk


ZhongpinWang (changes requested):
The lock file is generated.

Dependency flow: `@docusaurus/core` requires `webpack-dev-server` requires `selfsigned` requires `node-forge`.

deekshas8:
If node-forge has a vulnerability, the best way would be to always resolve it to a version which has the fix and higher. Manually editing the file is incorrect, as node-forge is a direct dependency of selfsigned.
Use overrides:

```
{
   "overrides": {
     "foo": "1.0.0"
   }
 }
```

deekshas8:
Looking at npm, there is no higher version for node-forge :/

CharlesDuboisSAP:
> Looking at npm, there is no higher version for node-forge :/

The last version was released 3 years ago, we should remove it.

PR #2067, opened 2025-04-14
Fix JS typedoc links

ZhongpinWang:
## What Has Changed?

Closes SAP/cloud-sdk-js#5717.

Due to a breaking change introduced in typedoc 0.27.0, the typedoc files are now generated with `-` instead of `_` for the package name. Thus, we need to change all the links.

Besides, some of the existing links in v4 were still pointed to v3 files.

All fixed in this PR.


tomfrenken:
Okay we missed a few (http_client -> http-client)

PR #2072, opened 2025-04-16
Fix broken links - Part: 2

tomfrenken:
## What Has Changed?


src/components/data/features.js
```diff
@@ -216,7 +216,7 @@ export const features = [
     },
     js: {
       status: YES,
-      docsLink: `[docs](${baseUrl}/js/guides/migrate-sdk-application-from-btp-cf-to-kubernetes)`,
+      docsLink: `[docs](${baseUrl}/js/environments/kubernetes)`,
```
ZhongpinWang, line 219: The other two links for `Environment` are also "broken" as the section does not exist any more in getting-started

ZhongpinWang (approved):
lgtm

PR #2124, opened 2025-06-13
chore: Remove csrf note

deekshas8:
## What Has Changed?


jjtang1985 (approved):
Thank you!

PR #2402, opened 2026-03-23
chore: [js] Document IAS token and destination helper functions

davidkna-sap:
## What Has Changed?


docs-js/features/connectivity/destination.mdx
```diff
@@ -275,6 +275,9 @@ Note, that if your `serviceBindingTransformFn()` function does not provide a nam
 
 More advanced examples with service token fetching can be found in [service-binding-to-destination.ts](https://github.com/SAP/cloud-sdk-js/blob/main/packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts).
 
+For the `identity` service type, the SAP Cloud SDK also provides the standalone convenience functions `getIasToken()` and `getIasDestination()`, which can also work with bare `ServiceCredentials` outside of a `VCAP_SERVICES` binding.
```
marikaner: [pp] Check the style guide. I think there are some slight violations of [this rule](https://github.com/SAP/cloud-sdk/blob/main/STYLEGUIDE.md#include-a-descriptive-noun-before-or-after-code).

docs-js/features/connectivity/ias.mdx
```diff
@@ -81,12 +81,16 @@ sequenceDiagram
 
 ### Creating Destinations
 
-Use [`getDestinationFromServiceBinding()`](pathname:///api/v4/functions/sap-cloud-sdk_connectivity.getDestinationFromServiceBinding.html) to connect to a system that is registered as an application within IAS.
+Use [`getDestinationFromServiceBinding()`](pathname:///api/v4/functions/sap-cloud-sdk_connectivity.getDestinationFromServiceBinding.html) or [`transformServiceBindingToDestination()`](pathname:///api/v4/functions/sap-cloud-sdk_connectivity.transformServiceBindingToDestination.html) to connect to a system that is registered as an application within IAS.
```
marikaner: [pp] I would prefer it, if we would better describe what happens here than "connect" because this might imply very different behavior. These functions do not connect to a system, they provide a destination that you can use to connect to a system.

docs-js/features/connectivity/ias.mdx
```diff
@@ -81,12 +81,16 @@ sequenceDiagram
...
 The parameter `iasOptions` contains:
 
 - `targetUrl`: The URL of the system where the target application resides.
 - `resource`: The dependency identified by its name or identifier configured in IAS (see [App2App Resources](#app2app-resources)) section.
 
+In addition to these standard functions for destination retrieval and transformation, the SAP Cloud SDK provides two convenience functions, [`getIasDestination()`](pathname:///api/v4/functions/sap-cloud-sdk_connectivity.getIasDestination.html) and [`getIasToken()`](pathname:///api/v4/functions/sap-cloud-sdk_connectivity.getIasToken.html), which fetch an IAS token and return a destination or token result respectively.
+These functions are useful when you need direct access to the IAS token or destination, for example outside SAP BTP environments with pre-populated `VCAP_SERVICES` or when constructing a destination manually.
+Refer to the [Convenience Functions](#convenience-functions) section below for more details.
```
marikaner: [pp] I am not 100% sure about the name `getIasDestination()`, because what is an IAS destination? Intuitively I would think it is a destination that uses IAS.
Consider [this file](https://github.com/SAP/cloud-sdk-js/blob/bbd9d92e99f349e5945056cce5318c4a648676b7/packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts), where we have all the functions to transform a service binding. 
1. I believe this function belongs in here.
2. Maybe we can align the naming. However, please do not use the naming as is right now in the file, because it violates the style guide. Instead how about `transformIasBindingToDestination()` or `createDestinationFromIasBinding()` or `createDestinationFromIasService()`.

marikaner: [req] While the point above is pp, I think it is more important to adjust the docs a little. Rough proposal (feel free to improve):

```suggestion
In addition to these functions for destination retrieval and transformation of service bindings, the SAP Cloud SDK provides two IAS-specific convenience functions:
- [`getIasToken()`](pathname:///api/v4/functions/sap-cloud-sdk_connectivity.getIasToken.html) to fetch an IAS token.
- [`getIasDestination()`](pathname:///api/v4/functions/sap-cloud-sdk_connectivity.getIasDestination.html) to transform the IAS binding to a destination with an IAS token.

These functions are useful when you need direct access to the IAS token or service, for example outside SAP BTP environments with pre-populated `VCAP_SERVICES` or when constructing a destination manually.
Refer to the [Convenience Functions](#convenience-functions) section below for more details.
```

marikaner: I am only now seeing the ias.mdx. This changes my latter comment. I would then suggest to just refer to this section and remove the duplicate information entirely.

docs-js/features/connectivity/ias.mdx
```diff
@@ -246,3 +245,73 @@ const destination = await getDestinationFromServiceBinding({
 });
 // Token request is automatically routed to the subscriber's IAS tenant
 ```
+
+## Convenience Functions
+
+The SAP Cloud SDK provides two convenience functions for working with IAS tokens directly.
+These are useful when you need access to the IAS token or destination, for example outside SAP BTP environments with pre-populated `VCAP_SERVICES` or when constructing a destination manually.
```
marikaner:
```suggestion
These are useful when you need access to the IAS token or service, for example outside SAP BTP environments with pre-populated `VCAP_SERVICES` or when constructing a destination manually.
```

docs-js/features/connectivity/ias.mdx
```diff
@@ -246,3 +245,73 @@ const destination = await getDestinationFromServiceBinding({
...
+
+Both functions accept either the string `'identity'` (preferred, resolves the binding from `VCAP_SERVICES`) or bare `ServiceCredentials` (e.g., just `clientid`, `clientsecret`, and `url`).
```
marikaner: [pp] Same as in the destination.mdx. Adhere to the style guide. I guess it would suffice to change it to "bare service credentials".

docs-js/features/connectivity/ias.mdx
```diff
@@ -246,3 +245,73 @@ const destination = await getDestinationFromServiceBinding({
...
+
+:::tip
+
+Pass `'identity'` whenever possible to let the SAP Cloud SDK resolve the IAS service binding from the environment, avoiding manual handling of credentials.
+
+:::
+
+- **[`getIasDestination()`](pathname:///api/v4/functions/sap-cloud-sdk_connectivity.getIasDestination.html)** fetches an IAS token and builds a ready-to-use [`Destination`](pathname:///api/v4/interfaces/sap-cloud-sdk_connectivity.Destination.html) with the token, the target URL, and the mTLS key pair from the service binding credentials (if present).
+- **[`getIasToken()`](pathname:///api/v4/functions/sap-cloud-sdk_connectivity.getIasToken.html)** fetches an IAS token and returns an [`IasTokenResult`](pathname:///api/v4/interfaces/sap-cloud-sdk_connectivity.IasTokenResult.html) with the access token string, its expiration, and an optional refresh token.
```
marikaner: [nth] Once more style guide returns an `IasTokenResult`. But only if you have a meaninful improvement in mind. I wouldn't be too strict about it.

docs-js/features/connectivity/ias.mdx
```diff
@@ -246,3 +245,73 @@ const destination = await getDestinationFromServiceBinding({
...
+
+The `targetUrl` is ignored if `getIasToken()` is used.
+
+:::
+
+```typescript
+import { getIasDestination, getIasToken } from '@sap-cloud-sdk/connectivity';
+
+// Use getIasDestination() to build a destination (technical user)
+// Preferred: pass 'identity' to resolve the binding from VCAP_SERVICES
```
marikaner: [req] We don't need to mention it so many times. I think the note above should be sufficient.

docs-js/features/connectivity/ias.mdx
```diff
@@ -246,3 +245,73 @@ const destination = await getDestinationFromServiceBinding({
...
+  {
+    targetUrl: 'https://backend-provider.example.com',
+    resource: { name: 'backend-api' }
+  }
+);
+```
+
+The `Destination` returned by `getIasDestination()` can be passed directly to any SAP Cloud SDK request builder or HTTP client.
+
+For the full set of options both functions accept the same [`IasTokenOptions`](pathname:///api/v4/types/sap-cloud-sdk_connectivity.IasTokenOptions.html) which includes `iasOptions` properties as available in [`getDestinationFromServiceBinding()`](pathname:///api/v4/functions/sap-cloud-sdk_connectivity.getDestinationFromServiceBinding.html) or [`transformServiceBindingToDestination()`](pathname:///api/v4/functions/sap-cloud-sdk_connectivity.transformServiceBindingToDestination.html).
```
marikaner: [req] Style guide, here I really think it makes sense to improve.
[req] I would put this right after the functions are introduced.

[pp] Do we really need this information? What value does it bring to the documentation. The API docs are linked and contain those details. My suggestion: why not remove this entirely?

marikaner (changes requested):

marikaner (approved):

PR #2415, opened 2026-04-07
Clarify caching behavior for SAMLAssertion destinations

marikaner:
Added note about SAMLAssertion authentication type not being cached.


KavithaSiva (approved):
LGTM

docs-js/features/connectivity/destination-cache-isolation.mdx
```diff
@@ -48,6 +48,11 @@ The destination cache holds destinations that potentially require HTTP requests,
 
 Registered destinations and destinations from environment variables are not held in the destination cache.
 
+:::note
+Destinations with authentication type "SAMLAssertion" are never cached.
```
KavithaSiva: [pp] Should we also add from which version this applies?

